PR #6, opened 2017-09-06
Add code to regenerate RECORD file for the wheel after we use patchelf

alykhantejani:
A fix for the broken `RECORD` files in the wheel (https://github.com/pytorch/pytorch/issues/2596)

This code should re-generate the web-safe SHA256 hashes and overwrite the current `RECORD` file

cc @soumith


soumith:
thanks a ton mate!!!

PR #17, opened 2017-10-23
Conda Install NNPACK for Jenkins builds

killeent:
Blocked by #16. When landed, I will publish the package to the PyTorch conda channel and verify that this works.


soumith:
landed #16, ball is in your court now.

PR #21, opened 2017-11-06
Add changes for ppc64le builds

avmgithub:
I'd like to submit changes to build_nimbix.sh for ppc64le builds.  I've spoken to Soumith about this.


soumith:
thank you!

PR #26, opened 2017-11-13
Smoke tests for examples/*

zou3519:
Doesn't include testing for examples/imagenet, that is a todo


test_community_repos/examples/dcgan/run-script.sh
```diff
@@ -0,0 +1,7 @@
+pushd examples/dcgan
+# smoke test
+python main.py --dataset fake --dataroot . --cuda --niter 1
```
soumith: can you switch this to 100 iterations, i'd like to catch things like memory leaks if possible. 100 iterations and still not OOM = maybe no memory leak :)

PR #44, opened 2018-03-19
Remove `pytorch` as a run-time dependency

esc:
This gives the user the flexibility to install either `pytorch` or `pytorch-cpu` depending on the hardware available. 

Without this, installing `torchvision` will forcefully install `pytorch` and all the associated `cuda` packages, thereby negating the advantages that `pytorch-cpu` has.

See also:  https://discuss.pytorch.org/t/possible-error-when-installing-pytorch-cpu-via-conda/15149

### DIFF
--- a/conda/torchvision-0.2.0/meta.yaml
+++ b/conda/torchvision-0.2.0/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - python
     - pillow >=4.1.1
     - numpy >=1.11
-    - pytorch >=0.3
     - six
 
 build:


soumith:
I wonder if a better solution is that I change the commands for CPU packages to have a `--skip-dep=pytorch` as mentioned in: https://github.com/conda/conda/issues/793#issuecomment-254475418

soumith:
what do you think?

esc:
Yeah, that would probably suffice too. It's a bit of a hack though.

Currently, the installation commands always reference both `pytorch` (or `pytorch-cpu`) and `torchvision`. Intuitively, this would indicate that `torchvision` and `pytorch` don't depend one each other in any way.
This is however currently not the case, as we have seen.

Looking at an example of the proposed command line:
```
conda install -c pytorch pytorch-cpu torchvision --skip-dep=pytorch
``` 

This looks quite silly and convoluted, to be honest and newbies would probably just execute the command without understanding what it does.

I think it might make more sense to get `pytorch` and `pytorch-cpu` to depend on `torchvision`. This would simplify the installation commands as follows:

```
conda install -c pytorch pytorch
```

and for CPU:

```
conda install -c pytorch pytorch-cpu
```

You could even use the long-form of `-c` i.e. `--channel` to disambiguate to make it more explicit that users are installing software from the `pytorch` channel. 

Anyway, last I should mention, that maybe there is some reason that I am not aware of, to include both the `pytorch` (or `pytorch-cpu`) and `torchvision` packages in the example installation commands.

esc:
@soumith any thoughs on this?

saareliad:
I encountered problem with this too when building a nightly version from source. (as version is written as "nightly").
If torchvision isn't a standalone package, then why does it make sense to require installation of its "parent" package?

PR #45, opened 2018-04-04
Update faiss build

ailzhang:
- faiss build with cuda80 and cuda91
- faiss-cpu not depend on cuda version
- remove cudatoolkit in faiss-gpu runtime.
nit: the missing newline was automatically removed by my editor... let me know if it's necessary. :P

### DIFF
--- a/conda/faiss-gpu/meta.yaml
+++ b/conda/faiss-gpu/meta.yaml
@@ -18,7 +18,6 @@ requirements:
     - python
     - numpy >=1.11
     - mkl
-    - cudatoolkit
 
 build:
   number: {{ environ.get('FAISS_BUILD_NUMBER') }}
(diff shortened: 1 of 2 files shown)


conda/faiss-gpu/meta.yaml
```diff
@@ -18,7 +18,6 @@ requirements:
     - python
     - numpy >=1.11
     - mkl
-    - cudatoolkit
```
soumith: why did you remove dependency on cudatoolkit?
are you assuming users has correct version of CUDA installed on system?

ailzhang: Mainly because we only have cudatoolkit8.0 in the channel. And if we want to install faiss-gpu with cuda90, it still installs cudatoolkit8.0 which is a bit weird.

PR #48, opened 2018-05-08
add copy of ignite master docs

alykhantejani:
cc @soumith I think this is all that's needed to get the ignite docs to display under pytorch.org/docs/master/ignite in a similar way to the vision ones at pytorch.org/docs/master/torchvision/


soumith:
closing as discussed (a long time ago)

PR #54, opened 2018-08-02
Don't run distributed tests for GPU builds

pjh5:
Or I can make this a flag or run only distributed with 'set +e'

### DIFF
--- a/manywheel/build.sh
+++ b/manywheel/build.sh
@@ -278,5 +278,5 @@ pushd $PYTORCH_DIR/test
 for PYDIR in "${DESIRED_PYTHON[@]}"; do
     "${PYDIR}/bin/pip" uninstall -y torch
     "${PYDIR}/bin/pip" install torch --no-index -f /$WHEELHOUSE_DIR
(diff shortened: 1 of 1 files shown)


manywheel/build.sh
```diff
@@ -278,5 +278,7 @@ pushd $PYTORCH_DIR/test
 for PYDIR in "${DESIRED_PYTHON[@]}"; do
     "${PYDIR}/bin/pip" uninstall -y torch
     "${PYDIR}/bin/pip" install torch --no-index -f /$WHEELHOUSE_DIR
-    LD_LIBRARY_PATH="/usr/local/nvidia/lib64" PYCMD=$PYDIR/bin/python $PYDIR/bin/python run_test.py
+    LD_LIBRARY_PATH="/usr/local/nvidia/lib64" PYCMD=$PYDIR/bin/python $PYDIR/bin/python run_test.py --exclude distributed
+    set +e
```
soumith: you need a `set -e` right after the test. or else the next loop run will ignore errors in other tests.

PR #59, opened 2018-08-06
Support for different package names in build*.sh

pjh5:
I will run a job to test that this works before removing wip

### DIFF
--- a/manywheel/build_cpu.sh
+++ b/manywheel/build_cpu.sh
@@ -208,10 +208,20 @@ cp /$WHEELHOUSE_DIR/torch*.whl /remote/$WHEELHOUSE_DIR/
 # remove stuff before testing
 rm -rf /opt/rh
 
+# The package's name that we made could be torch-nightly
+if [[ -n "$PACKAGE_NAME" ]]; then
+  package_name="$PACKAGE_NAME"
+else
+  package_name='torch'
+fi
+echo "Expecting the built wheels to be packages for '$package_name'"
+
+
+# Test that all the wheels work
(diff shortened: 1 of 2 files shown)


pjh5:
This works

manywheel/build_cpu.sh
```diff
@@ -208,10 +208,20 @@ cp /$WHEELHOUSE_DIR/torch*.whl /remote/$WHEELHOUSE_DIR/
...
+echo "Expecting the built wheels to be packages for '$package_name'"
+
+
+# Test that all the wheels work
 export OMP_NUM_THREADS=4 # on NUMA machines this takes too long
 pushd $PYTORCH_DIR/test
 for PYDIR in "${DESIRED_PYTHON[@]}"; do
-    "${PYDIR}/bin/pip" uninstall -y torch
-    "${PYDIR}/bin/pip" install torch --no-index -f /$WHEELHOUSE_DIR
+    "${PYDIR}/bin/pip" uninstall -y "$package_name"
```
soumith: this is super weird that this works (or is it not uninstalling maybe?). I think to allow a configurable package_name, you need to make changes to pytorch/setup.py, around here: https://github.com/pytorch/pytorch/blob/master/setup.py#L1026

pjh5: I have those corresponding changes here https://github.com/pytorch/pytorch/pull/9544 . I'm using that PR to test changes in the pytorch repo

pjh5: The environment variable that this expects PACKAGE_NAME is also written into the CI jenkins code

PR #68, opened 2018-08-22
Refactor build.sh and build_cpu.sh into a common script

anderspapitto:
carefully verified that everything was exactly the same between the
two scripts. Actually, I also found an unrelated bug - the cpu version
of the script used `PYDIR` rather than `pydir` in this snippet

    # Install the wheel for this Python version
    "${pydir}/bin/pip" uninstall -y "$package_name"
    "${pydir}/bin/pip" install "$package_name" --no-index -f /$WHEELHOUSE_DIR --no-dependencies

### DIFF
--- a/manywheel/build.sh
+++ b/manywheel/build.sh
@@ -2,14 +2,6 @@
 
 set -ex
 
-if [[ -z "$PYTORCH_BUILD_VERSION" ]]; then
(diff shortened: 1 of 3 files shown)


anderspapitto:
tested by https://ci.pytorch.org/jenkins/job/pip-packages/job/pip-cp27-cp27m-linux-build-upload/23/console

PR #70, opened 2018-08-22
source common script rather than run as subshell

anderspapitto:
some testing revealed that the previous approach breaks because
exporting arrays doesn't work in bash


anderspapitto:
build error fixed as verified in https://ci.pytorch.org/jenkins/job/pip-packages/job/pip-cp27-cp27m-cuda92-linux-build-upload/41/console

PR #90, opened 2018-08-31
Changes for cron job so far

pjh5:


cron/README
```diff
@@ -0,0 +1,5 @@
+# At its simplest you can do
+./prep_nightly.sh
```
orionr: Nice!

cron/build.sh
```diff
@@ -0,0 +1,154 @@
...
+        TORCH_PACKAGE_NAME='torch_nightly'
+    else
+        TORCH_PACKAGE_NAME='torch-nightly'
+    fi
+fi
+if [[ -z "$PYTORCH_BUILD_VERSION" ]]; then
+    PYTORCH_BUILD_VERSION='nightly'
+fi
+if [[ -z "$PYTORCH_BUILD_NUMBER" ]]; then
+    PYTORCH_BUILD_NUMBER='1'
```
orionr: Do you want this to be `1.0` instead of 1? Just checking.

pjh5: No the versions of these will be the date followed by '.post$PYTORCH_BUILD_NUMBER' if it's greater than 1. This is so we can have multiple builds per day

PR #146, opened 2018-09-25
Allowing individual uploads of builds

pjh5:


pjh5:
Merging now because it's easier to test and nightlies for the day have already completed

PR #191, opened 2018-11-06
Add nightly scripts for Windows CPU and CUDA90

peterjc123:


windows/internal/publish.bat
```diff
@@ -0,0 +1,49 @@
...
+
+:: Reset errorlevel
+ver >nul
+
+IF NOT EXIST %PACKAGE% mkdir %PACKAGE%
+
+xcopy /S /E /Y ..\..\output\*.* %PACKAGE%\
+
+git config --global user.name "Azure DevOps"
+git config --global user.email peterghost86@gmail.com
```
davidbrownellMS: Is there an automated account that can be used rather than an account tied to an individual?

windows/internal/test.bat
```diff
@@ -0,0 +1,39 @@
...
+
+for /F "delims=" %%i in ('where /R %SRC_DIR%\output\%CUDA_PREFIX% torch*%PYTHON_VERSION%*.whl') do pip install "%%i"
+
+if ERRORLEVEL 1 exit /b 1
+
+echo Smoke testing imports
+python -c "import torch"
+
+echo "Checking that MKL is available"
+python -c "import torch; exit(0 if torch.backends.mkl.is_available() else 1)"
```
davidbrownellMS: Will the python exit code be overwritten by subsequent calls?

windows/azure-pipelines.yml
```diff
@@ -0,0 +1,142 @@
...
+      CPU_3.6_Wheel:
+        DESIRED_PYTHON: 3.6
+      CPU_3.7_Wheel:
+        DESIRED_PYTHON: 3.7
+
+  pool:
+    vmImage: 'vs2017-win2016'
+
+  steps:
+  - task: BatchScript@1
```
davidbrownellMS: As written, batch files are invoked using the BatchScript@1 (e.g. line 29) task or the script task (e.g. line 51 and 54). It would be nice to use one format or the other.

windows/azure-pipelines.yml
```diff
@@ -0,0 +1,142 @@
+
```
davidbrownellMS: These changes look pretty good - nice job.

conda/pytorch-nightly/meta.yaml
```diff
@@ -9,6 +9,8 @@ requirements:
   build:
     - cmake
     - {{ compiler('c') }} # [win]
+
+  host:
```
pjh5: Is this change necessary for the Windows pipeline? Why's it needed?

peterjc123: According to the conda-build [docs](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host), once the ` {{ compiler() }}` jinja2 functions are used, there should be a host section. The reason that we use the aforementioned jinja2 function is that we have to override the default compiler. I just wonder why it doesn't work on Linux, so I'll try to add a Linux job first.

windows/azure-pipelines.yml
peterjc123: Since the script task doesn't support multiline scripts, I have to choose the first one. But I don't think it's nice to keep a one-line script in a separate file.

windows/internal/test.bat
peterjc123: Yes, I'll fix this.

windows/internal/publish.bat
peterjc123: No, I don't think I have one. But I can fake one if that's needed.

windows/internal/test.bat
peterjc123: Fixed!

conda/pytorch-nightly/meta.yaml
peterjc123: @pjh5 From the logs [here](https://peter-jc.visualstudio.com/pytorch_builder/_build/results?buildId=99&view=logs), it seems that the script works on Linux. Is there anything missing?

peterjc123: @pjh5 Plus: The config for the Linux job: https://github.com/pytorch/builder/pull/191/files#diff-f6392e6105a578ca4622abcfc34d2834R160.

pjh5: I just remember that this change broke the nightlies last time it was introduced. I'll run a job with these changes to make sure they still work

pjh5: Okay it seems that this works okay

windows/azure-pipelines.yml
peterjc123: Yes, I'll need these variables in the script. So it will be done after that issue is fixed.

peterjc123: Another question: Can I set additional environmental variables using BatchScript@1 just like that in the script task?
```yaml
  - task: BatchScript@1
    env:
      VARIABLE: VALUE
```

peterjc123:
@pjh5 Done!

pjh5:
LMK when you want to merge this.

peterjc123:
@pjh5 It's ready to be merged.

PR #193, opened 2018-11-06
Fixes to credentials during uploading

pjh5:


pjh5:
Seems like anaconda login issues are still present despite this, but I haven't been able to repo them consistently.

PR #271, opened 2019-02-21
Some changes for Windows

peterjc123:


mingwandroid:
Thanks @peterjc123, I believe that these DLLs can never have been loaded (or at least not in the recent past) as they're incompatible with newer nVidia DLLs in C:\Windows\System32.

Can you detail why they put in this folder in the first place? Did it fix something?

peterjc123:
@mingwandroid We put them in our binaries because of the following reasons. 
1. It doesn't break anything. As we all know, the DLLs in `C:\Windows\System32\` will always load first.
2. In some earlier versions like 0.4.0, users reported that if we put these DLLs in, they could use PyTorch without having to install CUDA.
I reverted it back now because none of these two points are true at present.

mingwandroid:
OK, thank you for the explanation @peterjc123.

PR #284, opened 2019-03-12
fix missing hashes from wheel's RECORD

bokhi:
a fix for the broken RECORD file in the wheel, the problem is similar to pytorch/pytorch#2596
which was already fixed in pytorch/builder/pull/6

basically since torch 1.0* the wheel also contains the cafee2/ folder but hashes are not generated for it


soumith:
thank you!

PR #292, opened 2019-04-24
Fix rpath in conda package.

beauby:
Fixes https://github.com/pytorch/pytorch/issues/19309.


beauby:
cc @fmassa

soumith:
thanks @beauby !

beauby:
Actually, the issue remains whenever `LD_LIBRARY_PATH` is explicitly set, which makes sense as it takes precedence over `DT_RUNPATH`, which `patchelf --set-rpath` sets by default (but not over `DT_RPATH`, which `patchelf --set-rpath` sets only if the `--force-rpath` option is provided). 
Conda build does set `DT_RPATH` (c.f. [https://github.com/conda/conda-build/blob/4d9dc1cb20dd02d70207a35ff2dd19e0b1efeca5/conda_build/post.py#L461](conda_build/post.py)). Should I issue a PR to add `--force-rpath`?

PR #297, opened 2019-04-29
Adding binary size backfill script and clearer comments

pjh5:
This is a script that used be running in jenkins which was forgotten in the migration to circleci. It is used for the binary sizes in the builder hud


pjh5:
cc @kostmo

PR #298, opened 2019-04-30
[Fix] Set DT_RPATH instead of DT_RUNPATH when patching conda libs.

beauby:
Follow up from #295/#296. CI in https://github.com/pytorch/pytorch/pull/19956 will make sure everything is ok.


soumith:
the test PR has CI failure that's relevant. commented on the PR

soumith:
thanks!

PR #299, opened 2019-04-30
Fixing parsing of cuda versions in binary size uploads

pjh5:
cc @kostmo


pjh5:
Looks to be working based off testing in https://github.com/pytorch/pytorch/pull/19934

PR #307, opened 2019-05-03
Remove dependency on version in binary_sizes_upload.sh

pjh5:


pjh5:
Tested with https://github.com/pytorch/pytorch/pull/20104 . Seems to work

PR #308, opened 2019-05-07
[Testing] Adding smoke test for gcc ABI

pjh5:
Not ready yet.

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -1,0 +1,192 @@
+#!/bin/bash
+set -eux -o pipefail
+
+# This script checks the following things on binaries
+# 1. The gcc abi matches DESIRED_DEVTOOLSET
+# 2. MacOS binaries do not link against OpenBLAS
+# 3. There are no protobuf symbols of any sort anywhere (turned off, because
+#    this is currently not true)
+# 4. Standard Python imports work
+# 5. MKL is available everywhere except for MacOS wheels
+# 6. CUDA is setup correctly and does not hang
+# 7. Magma is available for CUDA builds
+# 8. CuDNN is available for CUDA builds
+#
+# This script needs the env variables DESIRED_PYTHON, DESIRED_CUDA,
+# DESIRED_DEVTOOLSET and PACKAGE_TYPE
+#
+# This script expects PyTorch to be installed into the active Python (the
+# Python returned by `which python`). Or, if this is testing a libtorch
+# Pythonless binary, then it expects to be in the root folder of the unzipped
+# libtorch package. 
+
+
+# The install root depends on both the package type and the os
+# All MacOS packages use conda, even for the wheel packages.
+if [[ "$PACKAGE_TYPE" == libtorch ]]; then
+  install_root="$pwd"
+else
+  py_dot="${DESIRED_PYTHON:0:3}"
+  install_root="$(dirname $(which python))/../lib/python${py_dot}/site-packages/torch/"
+fi
+
+
+###############################################################################
+# Check GCC ABI
+###############################################################################
+echo "Checking that the gcc ABI is what we expect"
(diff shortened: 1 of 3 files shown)


check_binary.sh
```diff
@@ -0,0 +1,194 @@
...
+  if [[ "$(is_expected "$actual_gcc_abi")" != 1 ]]; then
+    echo "gcc ABI $actual_gcc_abi not as expected."
+    exit 1
+  fi
+
+  # We also check that there are[not] cxx11 symbols in libtorch
+  echo "Checking that symbols in libcaffe2.so have the right gcc abi"
+  libtorch="${install_root}/lib/libcaffe2.so"
+  set +e
+  cxx11_symbols="$(nm "$libtorch" | c++filt | grep cxx11)"
```
bwasti: grep for `__cxx11` rather than `cxx11`

check_binary.sh
```diff
@@ -0,0 +1,194 @@
...
+  has_cxx11_symbols="$?"
+  set -e
+  if [[ "$(is_expected "$has_cxx11_symbols")" != 1 ]]; then
+    if [[ "$has_cxx11_symbols" == 0 ]]; then
```
bwasti: I think this is backwards. Doesn't `$?` give the return code, which would be 0 when symbols are found

check_binary.sh
```diff
@@ -0,0 +1,194 @@
...
+  #   quote, any characters
+  #   Note the exactly one single character after the '='. In the case that the
+  #     variable is not set the '=' will be followed by a '"' immediately and the
+  #     line will fail the match and nothing will be printed; this is what we
+  #     want.  Otherwise it will capture the 0 or 1 after the '='.
+  # /.*D_GLIBCXX_USE_CXX11_ABI=\(.\)".*/
+  #   replace the matched line with the capture group and print
+  # /\1/p
+  actual_gcc_abi="$(sed -ne 's/.*D_GLIBCXX_USE_CXX11_ABI=\(.\)".*/\1/p' < "$torch_config")"
+  if [[ "$(is_expected "$actual_gcc_abi")" != 1 ]]; then
```
bwasti: I'm not a bash expert, but in general I think it is best practice to use 
```
"$(is_expected ${actual_gcc_abi})"
```

check_binary.sh
```diff
@@ -0,0 +1,194 @@
...
+python -c 'import torch'
+python -c 'from caffe2.python import core'
+
+
+###############################################################################
+# Check for MKL
+###############################################################################
+if [[ "$(uname)" != 'Darwin' || "$package_type" != *wheel ]]; then
+    echo "Checking that MKL is available"
+    python -c 'import torch; exit(0 if torch.backends.mkl.is_available() else 1)'
```
bwasti: `set -e`?

check_binary.sh
```diff
@@ -0,0 +1,194 @@
...
+
+###############################################################################
+# Check CUDA configured correctly
+###############################################################################
+# Skip these for Windows machines without GPUs
+if [[ "$OSTYPE" == "msys" ]]; then
+    GPUS=$(wmic path win32_VideoController get name)
+    if [[ ! "$GPUS" == *NVIDIA* ]]; then
+        echo "Skip CUDA tests for machines without a Nvidia GPU card"
+        exit 0
```
bwasti: perhaps just set `DESIRED_CUDA='cpu'`?

check_binary.sh
pjh5: Yes :(

check_binary.sh
pjh5: @peterjc123 has context on this. I only copied this from another file

pjh5:
Judging from https://github.com/pytorch/pytorch/pull/20507 , this actually looks safe to land now before the associated pytorch/pytorch changes

PR #314, opened 2019-06-14
Fix path after changing to out-source builds

peterjc123:
Should be merged after the change https://github.com/pytorch/pytorch/pull/21772 in PyTorch.


soumith:
cc: @pjh5

PR #316, opened 2019-06-19
Fix windows test logic

peterjc123:


windows/internal/test.bat
```diff
@@ -5,14 +5,20 @@ pushd %SRC_DIR%
 
 set PYTHON_VERSION=%PYTHON_PREFIX:py=cp%
 
+if "%MODULE_NAME%" == "pytorch" )
```
peterjc123: Looks like I missed a left bracket here.

PR #320, opened 2019-07-16
Enable switching to devtoolset7 for all nightlies

pjh5:
- Remove assumption that devtoolset7==new-gcc-abi
  - Renamed upgrade_gcc.sh to update_compiler.sh, to better match use
  - Fixed some logic in check_binary.sh to match devtoolset7 semantics
- Turn off checking libtorch symbols for cxx11, because the logic is wrong for devtoolset7
- Add back CUDA 9.2 support

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -37,14 +37,21 @@ fi
 echo "Checking that the gcc ABI is what we expect"
 if [[ "$(uname)" != 'Darwin' ]]; then
   function is_expected() {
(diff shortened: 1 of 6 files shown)


pjh5:
This should be merged in step with https://github.com/pytorch/pytorch/pull/22784

pjh5:
Tested with https://github.com/pytorch/pytorch/pull/23175

PR #328, opened 2019-07-31
Fix Windows LibTorch builds

peterjc123:
\+ means to concat files when using copy. Use quotes to see them as filenames.


peterjc123:
@soumith

PR #329, opened 2019-07-31
Get rid of package suffix; instead, use cpu-only feature to toggle CPU-ness.

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@fb.com>

### DIFF
--- a/conda/build_pytorch.sh
+++ b/conda/build_pytorch.sh
@@ -222,17 +222,13 @@ if [[ -n "$cpu_only" ]]; then
     if [[ "$OSTYPE" != "darwin"* ]]; then
         build_string_suffix="cpu_${build_string_suffix}"
     fi
-    # on Linux, rename the package pytorch-nightly-cpu, because it's cpu build
-    if [[ "$(uname)" != 'Darwin' ]]; then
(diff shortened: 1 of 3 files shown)


conda/build_pytorch.sh
```diff
@@ -222,17 +222,13 @@ if [[ -n "$cpu_only" ]]; then
...
         build_string_suffix="cpu_${build_string_suffix}"
     fi
-    # on Linux, rename the package pytorch-nightly-cpu, because it's cpu build
-    if [[ "$(uname)" != 'Darwin' ]]; then
-        export PYTORCH_PACKAGE_SUFFIX="-cpu" # used in name: pytorch part of meta.yaml
-    else
-        export PYTORCH_PACKAGE_SUFFIX=""
-    fi
+    # on Linux, advertise that the package provides cpu-only feature
+    export CONDA_CPU_ONLY_CONSTRAINT="    - cpu-only # [not osx]"
```
nairbv: how do we know it's not osx?

soumith: `conda-build` will take care of that. the comment `[not osx]` is a declarative thing for conda

PR #334, opened 2019-08-07
Add Ubuntu 16.04 docker image for building libtorch with new gcc ABI

yf225:
Copied most of the code from https://github.com/pytorch/builder/blob/master/conda/Dockerfile, and changed `yum` usage to `apt-get`.

Example docker image is at `yf225/tmp-binary-docker-image-ubuntu16.04:1`


ezyang:
I don't understand: if you're basing the image off of nvidia/cuda:9.2-devel-ubuntu16.04 why do you have to reinstall CUDA/CuDNN? They should come with the base image.

soumith:
@ezyang to be in control of the CUDA / CuDNN in a very exact way. The docker image might or might not have the version you want (some times, we want a previous CuDNN version or avoid a patch CUDA version because of known bugs)

ezyang:
In that case, I might ask the opposite question: if you're going to reinstall everything, why use NVIDIA's image at all? :)

ezyang:
I really don't like copy pasting here because at some point you're going to have to update the CUDA install and now there's two docker images you have to update to do this. It's not hard to write a bash script and include it in both docker files (see pytorch-dockerfiles for examples), can we use that?

common/install_cuda.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/bash
+
+set -ex
+
+# install CUDA 9.2 in the same container
+wget -q https://developer.nvidia.com/compute/cuda/9.2/Prod2/local_installers/cuda_9.2.148_396.37_linux -O setup
+chmod +x setup
+./setup --silent --no-opengl-libs --toolkit
+rm -f setup
```
yf225: In all of the scripts, I split `command a && command b` into `command a \n command b`, so that it's easier to see which line fails. The original script uses `&&` to concatenate multiple lines into one `RUN` command, likely to prevent Docker from creating too many layers (since it creates one layer per `RUN` command). It is not a concern now because we only run one `RUN` command per script.

common/install_cuda.sh
```diff
@@ -0,0 +1,147 @@
...
+ln -s libcudnn.so.7 cuda/lib64/libcudnn.so
+chmod +x cuda/lib64/*.so
+cp -a cuda/include/* /usr/local/cuda/include/
+cp -a cuda/lib64/* /usr/local/cuda/lib64/
+cd ..
+rm -rf tmp_cudnn
+ldconfig
+
+# # install CUDA 10.1 in the same container
+wget -q https://developer.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda_10.1.168_418.67_linux.run
```
yf225: The original download link `https://developer.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda_10.1.168_418.67_linux.` doesn't work, I changed it to the current one instead.

yf225:
I tried building `conda/Dockerfile` but it fails at `./cuda_10.1.168_418.67_linux.run --extract=/tmp/cuda`. Trying to figure out why.

PR #335, opened 2019-08-07
Build libtorch with new gcc ABI

yf225:
https://github.com/pytorch/pytorch/pull/23908


check_binary.sh
```diff
@@ -34,83 +35,104 @@ fi
...
-    #  fi
-    #else
-    #  if [[ -z "$1" || "$1" == 0 || "$1" == "OFF" ]]; then
-    #    echo 1
-    #  fi
-    #fi
-    if [[ -z "$1" || "$1" == 0 || "$1" == "OFF" ]]; then
-      echo 1
+
+# NOTE [ Building libtorch with old vs. new gcc ABI ]
```
ezyang: Any substantive changes here?

yf225: There is no substantive change here. The only main change is that we are only checking for cxx11 / pre-cxx11 symbols in our own namespaces (i.e. `c10` / `at` / `caffe2` / `torch`), since checking for those symbols in all namespaces doesn't give us reliable results:
```
  # NOTE: Checking the above symbols in all namespaces doesn't work, because
  # devtoolset7 always produces some cxx11 symbols even if we build with old ABI,
  # and CuDNN always has pre-cxx11 symbols even if we build with new ABI using gcc 5.4.
  # Instead, we *only* check the above symbols in the following namespaces:
```

PR #342, opened 2019-08-14
Add more binary tests for libtorch

yf225:
Currently our binary tests only test a very limited set of libtorch features. (For example, it doesn't test whether a simple include is successful, nor does it test existence of MKL / CUDA / CUDNN / MAGMA). This PR expands the binary test suite for libtorch.

Test PR: https://github.com/pytorch/pytorch/pull/24370


soumith:
great work, lgtm, thanks!

PR #349, opened 2019-09-13
Build macOS packages with torch.distributed support

pietern:
I figured these environment variables need to be set here, and the dependencies installed/managed here, so that if these scripts are used for the stable release builds, they carry over. If some other way is preferred, let me know. I tested this out at https://github.com/pytorch/pytorch/pull/26164 and the macOS conda and wheel builds are looking good (the CMake summary includes USE_DISTRIBUTED=1 and the builds succeed).

### DIFF
--- a/conda/build_pytorch.sh
+++ b/conda/build_pytorch.sh
@@ -114,6 +114,14 @@ else
         exit 1
     fi
 fi
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    # Produce macOS builds with torch.distributed support.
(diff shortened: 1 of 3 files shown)


pietern:
I wonder though -- there are many `meta.yaml` files lying around. Is `pytorch-nightly` also used (perhaps snapshotted as a copy) for the stable builds?

pietern:
@soumith Does this look good to you as well? Will this automatically carry over to the stable builds?

ezyang:
You're added. Please monitor binary builds in case of breakage.

PR #357, opened 2019-09-24
Add CUDA 3.0 to arch list

xsacha:
Now that 3.0 is working due to https://github.com/pytorch/pytorch/commit/d1496183f59b59928ac7003503591560716f21c5 is it possible to have the nightlies include this?

Thanks.


soumith:
no, 3.0 is too old, and our binary size is proportional to number of architectures we ship. We take a hard call that we are not shipping 3.0

PR #361, opened 2019-09-27
subprocess.run only support on python3

mingbowan:
tests are failing to due default python interpreter version is 2.


mingbowan:
well, python3 is not on default path, or maybe not installed at all :(

PR #376, opened 2019-10-16
Fix libiomp5.dylib not found issue for libtorch macOS binary

yf225:
This fixes https://github.com/pytorch/pytorch/issues/14727. The accompanying PR is https://github.com/pytorch/pytorch/pull/25208 which also adds tests to make sure we don't regress on this in the future.

### DIFF
--- a/wheel/build_wheel.sh
+++ b/wheel/build_wheel.sh
@@ -200,6 +200,7 @@ else
         cp -r "$(pwd)/any_wheel/torch/lib/include" "$(pwd)/libtorch/"
     fi
     cp -r "$(pwd)/any_wheel/torch/share/cmake" "$(pwd)/libtorch/share/"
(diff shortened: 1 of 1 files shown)


yf225:
Based on https://circleci.com/gh/pytorch/pytorch/3224767?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link, the test for checking MKL existence for macOS libtorch passes with this PR:
```cpp
#include <torch/torch.h>

int main(int argc, const char* argv[]) {
  TORCH_CHECK(torch::hasMKL(), "MKL is not available");
  return 0;
}
```
After merging this PR, I will update and merge https://github.com/pytorch/pytorch/pull/25208 to enable tests for macOS libtorch MKL.

yf225:
@soumith Would you like to review this PR? Thanks!

PR #383, opened 2019-11-14
add jni.h to docker100

IvanKobzarev:


dreiss:
Do we also need 101 and 92 for nightly and release builds?

soumith:
also added to the other docker files in 8b68132 and deployed the new images.

PR #386, opened 2019-11-25
Kick out cuda arch 3.5 for Windows

peterjc123:
@soumith @ezyang


xsacha:
> 3.7+PTX;5.0

Is there a reason this was changed to have +PTX on the 3.7 instead of 5.0?
Shouldn't PTX be generated on top of the latest supported feature set for optimal generation?

soumith:
you're right, because there's no arch between 3.7 or 5.0. But the +PTX is really adhoc protection, we shouldn't ever have user trigger that (it takes like 30 mins for PTX to SASS generation for PyTorch I think)

PR #400, opened 2020-01-31
Add CUDA 10.2 support

seemethere:
Kept CUDA 10.0 support for now until we can remove it from all the dependent CI systems.

Also refactored the `conda/Dockerfile` to be more parallelized using `DOCKER_BUILDKIT=1`

Will push `magma` and `pytorch/conda-cuda` once these are merged.


soumith:
all looks reasonable, but one thing we forgot to add back was that magma binaries have to be built with devtoolset3 in https://github.com/pytorch/builder/tree/master/conda#building-magma-cuda91

Basically, revert the lines 44 to 48 of README.md in https://github.com/pytorch/builder/commit/dd7edfdf40e7eb75f0d561512e31b31c8319d9c9

soumith:
or else, the binary size bug affects us

seemethere:
@soumith re-added those lines back to the readme, going to go ahead and merge.

PR #403, opened 2020-02-05
.circleci: Remove jobs that were not providing signal

seemethere:
These jobs were failing and taking a long time to fail so let's just
remove them for now and maybe add them back later.

It also might be worth it to remove the `nightly` jobs too since the `nightly-1` pipeline takes ~ 8hrs to run.


kostmo:
I'm OK with removing them for now.  Keep in mind these are the jobs that drive the [integration test results](http://ossci-integration-test-results.s3-website-us-east-1.amazonaws.com/test-results.html), so that when we re-instate integration tests (probably in the `pytorch` repo rather than the `builder` repo), we'll want to revive these.

PR #404, opened 2020-02-19
Add CUDA 10.2 support across the board

seemethere:
Have already pushed the docker image `pytorch/manylinux-cuda102`

### DIFF
--- a/common/install_magma.sh
+++ b/common/install_magma.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 function install_92 {
     # Install MAGMA for CUDA 9.2
     pushd /tmp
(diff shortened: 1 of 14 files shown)


peterjc123:
@seemethere Great! But it won't work for the nightly build machines until Microsoft installs CUDA 10.2 on their agents. If we drop CUDA 9.2 support, then we could just migrate the build jobs to CircleCI.

peterjc123:
CUDA 10.2 and cuDNN 7.6.5 are installed onto the Windows agents so you can trigger builds on them. BTW, you may need changes in https://github.com/pytorch/builder/blob/master/conda/pytorch-nightly/bld.bat#L25 for the conda builds.

PR #420, opened 2020-03-11
Update Windows CI for nightly builds

peterjc123:
According to https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent, `win1803` will be removed soon.


peterjc123:
cc @soumith @seemethere

PR #421, opened 2020-03-17
Add testing pipeline for promotion scripts

seemethere:


promote/wheel_to_pypi.sh
```diff
@@ -0,0 +1,46 @@
...
+# Dry run by default
+DRY_RUN=${DRY_RUN:-enabled}
+# On dry run just echo the commands that are meant to be run
+TWINE_UPLOAD="echo twine upload"
+if [[ $DRY_RUN = "disabled" ]]; then
+    TWINE_UPLOAD="twine upload"
+fi
+
+for pkg in ${pkgs_to_promote}; do
+    pkg_basename="$(basename "${pkg//linux/manylinux1}")"
```
ezyang: nit don't need double slash here right?

promote/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM debian as final
+RUN apt-get update && apt-get install -y curl unzip git ssh
```
ezyang: You should clear the apt cache when you're done to improve caching of the docker image

.circleci/config.in.yml
```diff
@@ -56,6 +56,23 @@ binary_common: &binary_common
     UNICODE_ABI: << parameters.unicode_abi >>
     CU_VERSION: << parameters.cu_version >>
 
+promote_common: &promote_common
+  docker:
+    - image: pytorch/release
```
ezyang: might be better to pin to a specific tag?

.circleci/config.in.yml
```diff
@@ -73,6 +90,37 @@ jobs:
...
             git diff --exit-code || (echo ".circleci/config.yml not in sync with config.yml.in! Run .circleci/regenerate.py to update config"; exit 1)
 
+  test_promote_s3:
+    <<: *promote_common
+    steps:
+      - checkout
+      - run:
+          name: Setting test environment variables
+          command: |
+            export PYTORCH_S3_FROM=nightly
```
ezyang: Does this actually work? I find it hard to believe that CircleCI is propagating environment variables across steps. Note that for example `binary_populate_env.sh` works by building a script which the final docker environment sources to set a bunch of envvars.

.circleci/config.in.yml
```diff
@@ -253,3 +301,19 @@ workflows:
     jobs:
       - circleci_consistency
       {{ workflows_nightly }}
+  test-promote:
+    # These are all dry-run by default
+    jobs:
+      # TODO: find a way to build the dockerfile per-pr
```
ezyang: We talked about priorities for what order to do things, but I think the proposed long docker build time should only take a ~day to prototype out in this context. Worth trying, I think!

promote/common_utils.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
+exit_if_not_on_git_tag() {
+    # Have an override for debugging purposes
+    if [[ -n "${TEST_WITHOUT_GIT_TAG-}" ]] ;then
```
ezyang: So, I don't personally write bash scripts with `set -u`, so this is new to me. However, it feels like slapping on the trailing hyphen on every variable usage is defeating the point of `set -u` (which is to tell you if you've, e.g., mispelled an env var). Wouldn't it be better to maintain the invariant that `TEST_WITHOUT_GIT_TAG` is always set in your scripts (by setting it if it is not set), and then using the non-hyphenated form?

promote/conda_to_conda.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
```
ezyang: I know bash script style is a very... ahem... personal matter. But I would suggest that we run `-x` on our scripts in CI (and in general, avoid suppressing output, e.g., piping to `/dev/null`). My personal reasoning for this, is at some point in the future I'm going to be reading the tea leaves for some build that failed in the past that I cannot conveniently rerun, and I want enough information to figure out what went wrong without having to go through another cycle of pushing and seeing what happened. The more information that's printed in the logs, the better for working out the cause of the problem post facto. These logs aren't going to be read by uninformed users, let's put as much info in them, and accept that maybe you have to scroll through some noise when diagnosing them.

promote/common_utils.sh
```diff
@@ -0,0 +1,76 @@
...
+        (
+            set -x
+            ${AWS} s3 cp ${DRY_RUN_FLAG} \
+                --only-show-errors \
+                --acl public-read \
+                "${PYTORCH_S3_BUCKET}/${from}" \
+                "${PYTORCH_S3_BUCKET}/${to}"
+        )
+    done < <(\
+        aws s3 ls --recursive "${PYTORCH_S3_BUCKET}/${package_type}/${PYTORCH_S3_FROM}" \
```
ezyang: Recursive ls makes my heart bleed (but I'm not sure there's another way unless you hardcode the flavors in this script itself)

promote/common_utils.sh
```diff
@@ -0,0 +1,76 @@
...
+            | grep -E "${package_name}-.*${pytorch_version}" \
+            | sed -e '/dev/d' \
+    )
+    # ^ We grep for package_name-.*pytorch_version to avoid any situations where domain libraries have
+    #   the same version on our S3 buckets
+}
```
ezyang: Don't you also need to update the s3 htmls?

promote/common_utils.sh
ezyang: Well, in a different PR, you use the S3 html index to determine what packages are available. Why not use that here?

PR #424, opened 2020-03-25
wheel: pin setuptools version to 46.0.0 on macOS

seemethere:
Relates to https://github.com/pytorch/pytorch/pull/35400

### DIFF
--- a/wheel/build_wheel.sh
+++ b/wheel/build_wheel.sh
@@ -137,10 +137,19 @@ export INSTALL_TEST=0 # dont install test binaries into site-packages
 export MACOSX_DEPLOYMENT_TARGET=10.10
 export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
 
+SETUPTOOLS_PINNED_VERSION=""
+PYYAML_PINNED_VERSION=""
+# These are currently pinned due to a bugged setuptools version,
+# see: https://github.com/pytorch/pytorch/pull/35400#issuecomment-604065952
+if [[ "${OSTYPE}" == "darwin"* ]]; then
+    SETUPTOOLS_PINNED_VERSION="=46.0.0"
+    PYYAML_PINNED_VERSION="=5.3"
+fi
+
 if [[ "$desired_python" == 3.8 ]]; then
-    retry conda install -yq cmake numpy=1.17 nomkl setuptools pyyaml ninja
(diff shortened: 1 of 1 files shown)


wheel/build_wheel.sh
```diff
@@ -137,10 +137,19 @@ export INSTALL_TEST=0 # dont install test binaries into site-packages
 export MACOSX_DEPLOYMENT_TARGET=10.10
 export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
 
+SETUPTOOLS_PINNED_VERSION=""
+PYYAML_PINNED_VERSION=""
+# These are currently pinned due to a bugged setuptools version,
+# see: https://github.com/pytorch/pytorch/pull/35400#issuecomment-604065952
+if [[ "${OSTYPE}" == "darwin"* ]]; then
+    SETUPTOOLS_PINNED_VERSION="=46.0.0"
+    PYYAML_PINNED_VERSION="5.3"
```
albanD: Is this missing an "=" ?

seemethere: Good catch, adding

seemethere:
> LGTM
> I guess you will run the tests on the binaries anyway and that will catch this as `torch_shm_manager` will fail to run but the particular thing to check herea are (from the root of the source code):
> 
> * `torch/bin/` should be fine
> * `build/bin/` should be fine
> * `build/lib.macosx-10.14-x86_64-3.7/torch/bin/` is the one that was faulty. The content should have been copied from `torch/bin/` as is but the x bit was dropped. The install takes the bin from this folder with the latest setuptools version.

Yup looks like after building with this script we are good to go:
```
❯ ls -la $(find . -name "torch_shm_manager")
-rwxr-xr-x  1 eliuriegas  staff  43228 Mar 25 15:50 ./tmp_wheel_conda_3.8_151518/pytorch/build/bin/torch_shm_manager
-rwxr-xr-x  1 eliuriegas  staff  43228 Mar 25 15:50 ./tmp_wheel_conda_3.8_151518/pytorch/build/lib.macosx-10.9-x86_64-3.8/torch/bin/torch_shm_manager
-rwxr-xr-x  1 eliuriegas  staff  43228 Mar 25 15:50 ./tmp_wheel_conda_3.8_151518/pytorch/torch/bin/torch_shm_manager

❯ ls -la ${CONDA_PREFIX}/lib/python3.8/site-packages/torch/bin/torch_shm_manager

-rwxr-xr-x  1 eliuriegas  staff  43228 Mar 25 15:53 /Users/eliuriegas/.pyenv/versions/miniconda3-4.3.30/envs/test38/lib/python3.8/site-packages/torch/bin/torch_shm_manage
```

PR #436, opened 2020-04-14
Copy hidden files for PyTorch wheel builds

peterjc123:
(e.g. `.git`) otherwise subsequent git calls won't work.
cc @seemethere @soumith


peterjc123:
@seemethere and this one, please

PR #444, opened 2020-05-18
so people don't try to run the commands in the container's bash

jayenashar:


manywheel/README.md
```diff
@@ -13,7 +13,7 @@ OR
 nvidia-docker run -it --ipc=host --rm -v $(pwd):/remote soumith/manylinux-cuda100:latest bash
 ```
 
-Then run:
+Exit the container, then back in the host, run:
```
antoniovs1029, line 16: @jayenashar I know this is an old PR, but I'm trying to follow the instructions and the logic, and don't understand... if I run the `nvidia-ducker run -it` above, and then "exit the container", what's the point of running the docker run? since it isn't detached, exiting it will simply stop the container... furthermore, it seems that the logic in here actually starts another container anyway, so I'm confused about this

https://github.com/pytorch/builder/blob/52c2f25f20164f1d4d36c620c451a6577353637c/cron/build_docker.sh#L120-L127

perhaps the instructions are outdated?

jayenashar: i can't recall and i've given up on building pytorch packages, but i would guess it's to pre-download the image.  i didn't write lines 11-13 so i really don't know.  you may also notice the `--rm` which will remove the stopped container.

PR #446, opened 2020-05-27
Don't install VS2019

peterjc123:
To make binary jobs work in the new circleci image.


peterjc123:
cc @soumith @seemethere @malfet

PR #448, opened 2020-05-29
Add randomtemp to build scripts

peterjc123:
Fixes https://github.com/pytorch/pytorch/issues/25393.
Applies https://github.com/pytorch/pytorch/pull/35375/files to pytorch/builder.
@malfet @seemethere @soumith

### DIFF
--- a/conda/pytorch-nightly/bld.bat
+++ b/conda/pytorch-nightly/bld.bat
@@ -65,6 +65,20 @@ if "%CUDA_VERSION%" == "80" (
     set "CUDAHOSTCXX=%VS140COMNTOOLS%\..\..\VC\bin\amd64\cl.exe"
 )
 
+:: randomtemp is used to resolve the intermittent build error related to CUDA.
+:: code: https://github.com/peterjc123/randomtemp
+:: issue: https://github.com/pytorch/pytorch/issues/25393
+::
+:: Previously, CMake uses CUDA_NVCC_EXECUTABLE for finding nvcc and then
+:: the calls are redirected to sccache. sccache looks for the actual nvcc
+:: in PATH, and then pass the arguments to it.
+:: Currently, randomtemp is placed before sccache (%TMP_DIR_WIN%\bin\nvcc)
+:: so we are actually pretending sccache instead of nvcc itself.
+curl -kL https://github.com/peterjc123/randomtemp/releases/download/v0.3/randomtemp.exe --output %SRC_DIR%\tmp_bin\randomtemp.exe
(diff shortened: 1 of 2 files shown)


conda/pytorch-nightly/bld.bat
```diff
@@ -65,6 +65,20 @@ if "%CUDA_VERSION%" == "80" (
...
+:: issue: https://github.com/pytorch/pytorch/issues/25393
+::
+:: Previously, CMake uses CUDA_NVCC_EXECUTABLE for finding nvcc and then
+:: the calls are redirected to sccache. sccache looks for the actual nvcc
+:: in PATH, and then pass the arguments to it.
+:: Currently, randomtemp is placed before sccache (%TMP_DIR_WIN%\bin\nvcc)
+:: so we are actually pretending sccache instead of nvcc itself.
+curl -kL https://github.com/peterjc123/randomtemp/releases/download/v0.3/randomtemp.exe --output %SRC_DIR%\tmp_bin\randomtemp.exe
+set RANDOMTEMP_EXECUTABLE=%SRC_DIR%\tmp_bin\nvcc.exe
+set CUDA_NVCC_EXECUTABLE=%SRC_DIR%\tmp_bin\randomtemp.exe
```
malfet, line 79: Shouldn't this be other way around?

peterjc123: Could you please give an example?

peterjc123: Currently, we can only workground this issue by retrying it several times.

peterjc123:
The cuda builds in nightly jobs seem to be more flaky with the new circleci image which has VS 16.6 installed. So I hope that we could land this asap.

conda/pytorch-nightly/bld.bat
malfet: Why `RANDOMTEMP_EXECUTABLE` is points to `nvcc` and `CUDA_NVCC_EXECUTABLE` poitns to `randomtemp.exe`?

peterjc123: Because `randomtemp.exe` acts like a shim exectuable, it just passes the argument to the actual program, which is specified by the variable `RANDOMTEMP_EXECUTABLE`.

seemethere: Yea I agree with @malfet this does seem a bit confusing when you look at it.

Maybe it'd be better to add a comment here explaining why they point where they point to.

malfet: I think I understand how it works, but naming of variables is a bit confusing, but indeed it matches one on master

PR #462, opened 2020-06-18
Clean up old CUDA support for Windows

zasdfgbnm:
cc: @peterjc123


zasdfgbnm:
@peterjc123 Now it is deleted

PR #464, opened 2020-06-30
Update VS 2017 C++ toolchain to 14.13

peterjc123:
Testing with https://github.com/pytorch/pytorch/pull/40782.


peterjc123:
cc @seemethere @soumith @malfet

PR #467, opened 2020-07-16
PR: Add FFmpeg recipe to builder

andfoy:
This PR adds ffmpeg to the pytorch conda channel in order to be used by torchvision. GPL and static components were disabled in order to prevent licensing issues.

cc @fmassa, @seemethere


andfoy:
I tested this PR locally on Linux, and it works as expected

andfoy:
I've also tested locally on Windows, and it also works

seemethere:
I can build the packages for these for linux, but I don't think I have a windows machine on hand

andfoy:
> I can build the packages for these for linux, but I don't think I have a windows machine on hand

@seemethere  I have a Windows VM, please let me know if that works

PR #470, opened 2020-07-25
Use MSVC 14.11 for 1.6 stable binaries

peterjc123:
Fix the test failure in jit.
@seemethere @soumith @malfet


malfet:
I'm fine with the change, but:
 - Hopefully we are not going to rebuild 1.6 for Windows with CUDA-9.2
 - We should apply this change to master as well, as the same problem likely exists there as well.
 - Although it might fix the problem, it does not explain why the problem happens in the first place (is there a regression in 14.13 than forces us to use 14.11?)

peterjc123:
> * We should apply this change to master as well, as the same problem likely exists there as well.

According to @guyang3532 's test, this problem doesn't exist on master.

PR #476, opened 2020-08-04
manywheel: Move to manylinux2014

seemethere:

### DIFF
--- a/manywheel/Dockerfile
+++ b/manywheel/Dockerfile
@@ -1,17 +1,44 @@
 # syntax = docker/dockerfile:experimental
 ARG BASE_CUDA_VERSION=10.1
-FROM centos:7 as base
+FROM quay.io/pypa/manylinux2014_x86_64 as python
+# remove unncessary python versions
+RUN rm -rf /opt/python/cp27* /opt/_internal/cpython-2.7*
+RUN rm -rf /opt/python/cp35* /opt/_internal/cpython-3.5*
 
+FROM python as libcrypt
+RUN mkdir -p /libcrypt && find /usr/local/lib -type f -name "libcrypt*" -exec cp -L {} /libcrypt \;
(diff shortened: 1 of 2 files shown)


manywheel/Dockerfile
```diff
@@ -1,17 +1,44 @@
...
+        perl \
+        unzip \
+        util-linux \
+        wget \
+        which \
+        xz \
+        yasm
 RUN yum install -y yum-utils centos-release-scl
 RUN yum-config-manager --enable rhel-server-rhscl-7-rpms
 RUN yum install -y devtoolset-7-gcc devtoolset-7-gcc-c++ devtoolset-7-gcc-gfortran devtoolset-7-binutils
```
seemethere, line 39: We should also probably have a discussion on whether or not we want to use devtoolset-7 or devtoolset-9, which is what manylinux2014 upstream uses 

https://github.com/pypa/manylinux/blob/bb4aa00ea89acf22d3b35c6d909660da84533b31/docker/build_scripts/build.sh#L63

soumith: dont care as long as it compiles. afaicr devtoolset9 (gcc-9) didn't compile pytorch when we checked a long time ago. it probably does, now.

seemethere: This is actually still the case, the highest version of gcc that CUDA 10.1 supports is still `8.x` :(

seemethere: Do you think it'd be an issue if the version of python we compile against was compiled with devtoolset-9 while we still compile with devtoolset-7?

mattip: I think it is fine. Linux does not suffer from the same runtime-mixing problems that windows has.

janaknat:
Will this PR be accepted? We are looking at add support for pytorch aarch64 wheels. That will require manylinux2014.

seemethere:
I'm currently re-doing this within the pytorch repository here: https://github.com/pytorch/pytorch/pull/45091

PR #478, opened 2020-08-04
PR: Add FFmpeg conda release pipeline

andfoy:
This PR adds a CircleCI workflow to build and upload FFmpeg distributions to the PyTorch conda channel.

cc @seemethere @fmassa

### DIFF
--- a/.circleci/config.in.yml
+++ b/.circleci/config.in.yml
@@ -317,9 +317,21 @@ workflows:
 
   ffmpeg:
     jobs:
-      - ffmpeg_linux
-      - ffmpeg_mac
-      - ffmpeg_win
+      - ffmpeg_linux:
+          cu_version: cpu
+          is-python3: true
+          name: ffmpeg
+          python_version: '3.7'
+      - ffmpeg_mac:
+          cu_version: cpu
+          is-python3: true
+          name: ffmpeg
+          python_version: '3.7'
+      - ffmpeg_win:
+          cu_version: cpu
+          is-python3: true
+          name: ffmpeg
+          python_version: '3.7'
       - hold:
           type: approval
           requires:
--- a/.circleci/config.in.yml
+++ b/.circleci/config.in.yml
@@ -280,7 +280,7 @@ jobs:
       - run:
           name: Build FFmpeg recipe
           command: |
-            ffmpeg/build_ci.sh
+            .circleci/helper-scripts/wrap-with-conda-activation-macos.sh ffmpeg/build_ci.sh
       - persist_to_workspace:
           root: ffmpeg
           paths:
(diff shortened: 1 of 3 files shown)


.circleci/config.in.yml
```diff
@@ -251,5 +286,16 @@ workflows:
               only:
                 - master
     jobs:
+      - hold:
```
seemethere, line 289: It'd probably be good to make this into it's own separate workflow, so that it can also be triggered on PRs

ffmpeg/build_ci.sh
```diff
@@ -0,0 +1,19 @@
...
+
+LATEST_COMMIT=$(git rev-parse HEAD)
+FOLDER_COMMIT=$(git log -1 --format=format:%H --full-diff $DIR)
+
+if [ $FOLDER_COMMIT = $LATEST_COMMIT ]; then
+    echo "FFMpeg has changed"
+    ./$DIR/build_ffmpeg.sh
+
+    echo "Uploading FFmpeg to PyTorch Anaconda channel"
+    anaconda upload -u pytorch --force $DIR/output/*/ffmpeg*.bz2
```
seemethere, line 15: We should probably just split this out into it's own job so we can have separate build / upload jobs.

This is to make it easier to test any changes as they come in.

ffmpeg/build_ci.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+LATEST_COMMIT=$(git rev-parse HEAD)
+FOLDER_COMMIT=$(git log -1 --format=format:%H --full-diff $DIR)
+
+if [ $FOLDER_COMMIT = $LATEST_COMMIT ]; then
```
seemethere, line 10: nice :+1:

ffmpeg/build_ci.sh
seemethere: You can persist workspace artifiacts using the `persist_to_workspace` functionality within circleci: https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace

.circleci/config.in.yml
andfoy: Do I need to add something else here?

andfoy:
@seemethere, do I need to execute ``python regenerate.py``?

fmassa:
@andfoy I don't know about `builder`, but I would expect so, given that CircleCI will only pick the `config.yaml` file (which is autogenerated). So I would say you'd need to execute this command and commit `config.yaml` as well

andfoy:
@seemethere, this is ready for approval!

PR #487, opened 2020-08-11
Fix CUDA-10.2 installation logic on Windows

malfet:
Add missing `goto cuda_common` otherwise installing 10.2 will actually install CUDA-11


seemethere:
Yup I'm fixing the CI for this one currently in https://github.com/pytorch/builder/pull/486

PR #491, opened 2020-08-13
llvm: Add llvm docker build

seemethere:
To be used by upstream docker images, builds took too long in upstream


soumith:
why are we using llvm?

seemethere:
It was originally used for some NNC tests although I'm not sure I have the full context on why it's actually necessary.

https://github.com/pytorch/pytorch/pull/35564

PR #500, opened 2020-08-25
manywheel: Add arch 8.0 to CUDA 11.0 builds

seemethere:


soumith:
you mean to CUDA 11.0 builds

seemethere:
Oh you're right

PR #501, opened 2020-08-25
common: Comment out pruning for CUDA 11

seemethere:
We ran into issues allocating memory with pruned CUDA dependencies, so
we're going to forgo pruning for now

### DIFF
--- a/common/install_cuda.sh
+++ b/common/install_cuda.sh
@@ -223,14 +223,14 @@ function prune_110 {
 
 
     # all CUDA libs except CuDNN and CuBLAS (cudnn and cublas need arch 3.7 included)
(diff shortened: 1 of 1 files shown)


soumith:
cc: @zasdfgbnm @csarofeen 

if we can't prune, we can't ship CUDA 11 binaries to PyPI, and I know CuBLAS/CuDNN releases are really long, so worth getting a head-start on this

malfet:
https://github.com/pytorch/builder/pull/502/ fixes the pruning problem by updating to the latest CUDA-11 toolkit version, so I guess this PR can be safely abandoned

seemethere:
Closing since #502 solves our issues

PR #517, opened 2020-09-11
add ability to copy auxiliary deps into wheel

jeffdaily:
This is needed to support rocblas tensile.

CC @sunway513


jeffdaily:
@walterddr please review.  When testing our nightly rocm wheels, we ran into a runtime failure where our rocblas library expected some auxiliary files to be present in the same directory.  I've added an ability to copy arbitrary files into the wheel.  Would be a no-op on any other platform, but potentially useful outside of our ROCm needs.

walterddr (approved):
@seemethere can you add me with write access to the builder repo. seems like I am not allow to merge changes

seemethere (approved):

PR #521, opened 2020-09-17
Add promotion script for pytorch releases

seemethere:

### DIFF
--- a/release/promote.sh
+++ b/release/promote.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+
+# Make sure to update these versions when doing a release first
+PYTORCH_VERSION=${PYTORCH_VERSION:-1.6.0}
+TORCHVISION_VERSION=${TORCHVISION_VERSION:-0.7.0}
+TORCHAUDIO_VERSION=${TORCHAUDIO_VERSION:-0.6.0}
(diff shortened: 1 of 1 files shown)


release/promote.sh
```diff
@@ -0,0 +1,95 @@
...
+promote_conda pytorch conda "${PYTORCH_VERSION}"
+promote_conda torchvision conda "${TORCHVISION_VERSION}"
+promote_conda torchaudio conda "${TORCHAUDIO_VERSION}"
+promote_conda torchtext conda "${TORCHTEXT_VERSION}"
+promote_conda torchserve conda "${TORCHSERVE_VERSION}"
+promote_conda torch-model-archiver conda "${TORCHSERVE_VERSION}"
+promote_conda torchcsprng conda "${TORCHCSPRNG_VERSION}"
+
+# Uncomment these to promote to pypi
+# promote_pypi torch "${PYTORCH_VERSION}"
```
walterddr, line 91: would this script be used inside `pytorch/pytorch/.circleci`? I was wondering if we can just use env var to control this?

walterddr (approved):
left one question, but overall lgtm

release/promote.sh
scopatz: Or maybe a command line argument

seemethere: No we typically promotion is the last manual step in the release process.

As part of this, we make it purposefully difficult to promote to PyPI since it cannot be undone.

PR #524, opened 2020-09-18
PR: Build FFmpeg 4.2

andfoy:
This PR adds FFmpeg 4.2 builds to the PyTorch channel, it also disables packages for Windows temporarily.
See https://github.com/pytorch/vision/issues/2650 and https://github.com/pytorch/vision/pull/2596#issuecomment-695030369

cc @seemethere @fmassa

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,7 @@ binary_common: &binary_common
     PYTORCH_VERSION: << parameters.pytorch_version >>
     UNICODE_ABI: << parameters.unicode_abi >>
     CU_VERSION: << parameters.cu_version >>
+    TERM: xterm-256color
 
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,17 +148,11 @@ workflows:
           is-python3: true
           name: ffmpeg_mac
           python_version: '3.7'
-      - ffmpeg_win:
-          cu_version: cpu
-          is-python3: true
-          name: ffmpeg_win
-          python_version: '3.7'
       - hold:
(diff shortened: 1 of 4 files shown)


ffmpeg/build_ci.sh
```diff
@@ -20,14 +20,22 @@ git fetch origin
...
+COMMIT_SAME=$?
+
 set -exou pipefail
 
 mkdir -p "${DIR}/output/"
 # This is so that circle's persist_to_workspace doesn't fail when builds don't run
 touch "${DIR}/output/hello"
 
-if git merge-base --is-ancestor "${FOLDER_COMMIT}" "${BASE_COMMIT}"; then
+if [ $COMMIT_SAME -eq 1 ]; then
```
seemethere, line 32: The previous command should just work, I don't think this change is actually necessary

.circleci/config.yml
```diff
@@ -148,17 +142,11 @@ workflows:
           is-python3: true
           name: ffmpeg_mac
           python_version: '3.7'
-      - ffmpeg_win:
```
seemethere, line 151: Should we be removing windows here?

ffmpeg/build_ci.sh
andfoy: I tried to run it both locally and on the CI on the previous version, and it didn't work at all, I think it is related to the `set -exou` that comes after assigning the variable

seemethere: I definitely thought conditionals were exempt from the `set -e` erroring out, but I'm willing to just push this through since it's relatively small.

Issue #504, opened 2020-08-30
Update to recent MKL version and Zen performance


danieldk:
This is a follow-up to my comments in #460, but I thought it would be better to make an issue out of this rather adding more comments to a merged pull request ;).

The question in short: `MKL_DEBUG_CPU_TYPE` cannot be used anymore in recent BLAS versions. Is it still possible to use AVX2-optimized kernels on AMD Zen CPUs?

I did some more investigation. The good news is that apparently Intel is integrating Zen support in MKL. The bad news is that it is that Zen kernels haven't been implemented for every BLAS function, and if no Zen function is available, it switches to the slow SSE kernel.

The following is using MKL 2020.2.254 and a Ryzen 3700X

First, I use the standard ACES DGEMM benchmark:

```
$ ./mt-dgemm 4000 | grep GF
GFLOP/s rate:         227.809737 GF/s
```

Hotspot in `perf`:

```
65.34%  mt-dgemm  libmkl_def.so       [.] mkl_blas_def_dgemm_kernel_zen
```

Clearly a Zen-optimized kernel. Then I made an SGEMM version of the same benchmark:

```
$ ./mt-sgemm 4000 | grep GF
GFLOP/s rate:         151.946679 GF/s
```

Not so stellar. `perf` reveals code path using SSE (I checked the instructions used):

```
74.26%  mt-sgemm  libmkl_def.so     [.] LM_LOOPgas_1
```

Next, we use `LD_PRELOAD` to override the function that detects Intel CPUs to always return true:

```
$ LD_PRELOAD=libfakeintel.so ./mt-sgemm 2000 | grep GF
GFLOP/s rate:         382.358381 GF/s
```

Much better! Top function in `perf`:

```
59.73%  mt-sgemm  libmkl_avx2.so          [.] mkl_blas_avx2_sgemm_kernel_0
```

tl;dr: MKL seems to be moving towards supporting AMD Zen. However, it seems that Zen kernels haven't been implemented for every BLAS function yet. Possible (hopefully) temporary workaround: put `DT_NEEDED` item in a library or program's ELF dynamic section to override detection.

malfet:
Can you please run the speed comparison between MKL 2020.0 (last version that still supports MKL_DEBUG_CPU_TYPE) and 2020.3, which should have some perf regression fixed? (see https://software.intel.com/content/www/us/en/develop/articles/intel-math-kernel-library-release-notes-and-new-features.html )

danieldk:
[...]

PR #527, opened 2020-09-22
Downgrade MKL to 2020.0

malfet:
According to community reports it is the last version of MKL that runs AVX2-optimized kernels on AMD Zen CPUs.

Fixes https://github.com/pytorch/builder/issues/504

### DIFF
--- a/common/install_mkl.sh
+++ b/common/install_mkl.sh
@@ -3,8 +3,8 @@
 set -ex
 
 # MKL
-MKL_VERSION=2020.1
-MKL_BUILD=217
+MKL_VERSION=2020.0
+MKL_BUILD=166
 mkdir -p /opt/intel/lib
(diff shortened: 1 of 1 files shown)


soumith (approved):
nice!

seemethere (approved):

seemethere:
Oh wait does this mean we have to re-build all of the docker images o.O

soumith:
This is only for manywheel right?

wrt conda, we currently have `mkl >= 2018`. We might not be able to change it to  add the `< 2020.1` because that would downgrade other conda packages down, such as numpy/scipy. realized this, so wanted to flag before I forgot.

seemethere:
Alright, going to re-build all of the manywheel images today then, will update here when they've been pushed.

PR #528, opened 2020-09-22
add ROCm 3.8 manywheel build

jeffdaily:

PR #531, opened 2020-09-24
[hotfix] follow up changes to rocm3.8

walterddr:
follow up on #528

### DIFF
--- a/cron/update_s3_htmls.sh
+++ b/cron/update_s3_htmls.sh
@@ -9,7 +9,7 @@ set -e
 
 # Upload for all CUDA/cpu versions if not given one to use
 if [[ -z "$CUDA_VERSIONS" ]]; then
-    export CUDA_VERSIONS=('cpu' 'cu92' 'cu100' 'cu101' 'cu102' 'cu110' 'rocm3.7')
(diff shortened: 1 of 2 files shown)


walterddr:
@jeffdaily @seemethere

cron/update_s3_htmls.sh
```diff
@@ -9,7 +9,7 @@ set -e
 
 # Upload for all CUDA/cpu versions if not given one to use
 if [[ -z "$CUDA_VERSIONS" ]]; then
-    export CUDA_VERSIONS=('cpu' 'cu92' 'cu100' 'cu101' 'cu102' 'cu110' 'rocm3.7')
+    export CUDA_VERSIONS=('cpu' 'cu92' 'cu100' 'cu101' 'cu102' 'cu110' 'rocm3.7' 'rocm3.8')
```
walterddr, line 12: @seemethere should we add this together with the nightly PR or should we add it AFTER the files are being uploaded?

seemethere: it doesn't actually matter, this will work even if there are no `rocm3.8` packages available

seemethere (approved):

PR #532, opened 2020-09-25
@jeremiahschung Fix date handling in download_count_wheels.py

jeremiahschung:
The existing script called key.last_modified.utcnow() which always returned the current time regardless of the value in the datetime object last_modified. This resulted in the script downloading all logs every time (the continue statement never executed due to miscalculation in the if statement). Fixed it to pull logs from the previous day only.

Test result:
Run successfully finished while downloading only files from yesterday.
[download_count_wheels_test_run.log](https://github.com/pytorch/builder/files/5285581/download_count_wheels_test_run.log)

### DIFF
--- a/analytics/download_count_wheels.py
+++ b/analytics/download_count_wheels.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 import gzip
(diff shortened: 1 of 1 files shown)


walterddr (approved):
lgtm. @seemethere can you take another look?

seemethere (approved):

PR #540, opened 2020-10-06
Update download_count_wheels.py

jeremiahschung:
Due to the `entries` variable being scoped outside of the for loop that processes a single file, the counts are dramatically inflated (in my case processing data for 1 month, the incorrect calculation was 230x larger than the actual number). Entries in the first logfile gets counted N times where N is the number of files to process.


jeremiahschung:
@walterdd @seemethere Please take a look!

seemethere (approved):

PR #548, opened 2020-10-10
include dataclasses in 3.6

walterddr:


walterddr:
include dataclasses in run_test is not the right way to go. it should be packaged with torch

PR #551, opened 2020-10-12
include dataclasses in py36 for run_test and smoke_test

walterddr:
Fix installation issue for dataclasses when running tests.


seemethere (commented):
Would this also work for windows since windows uses `windows/internal/smoke_test.bat`?

walterddr:
looks like windows conda is using build_pytorch.sh (then run_test.sh) see: https://github.com/pytorch/builder/blob/master/windows/internal/build_conda.bat yes?

walterddr:
oops. you are right. this is next - once binary_windows_conda_xxx build passes, it needs to run smoke_test.bat... didn't think that far. good catch!!!

walterddr:
closing in favor of #552

PR #552, opened 2020-10-12
fix dataclasses installation in smoke_test and run_test

walterddr:
create a new one for easier testing on pytorch/pytorch


seemethere (approved):

walterddr:
`check_binary.sh` is still broken :-( will prepare another fix.

PR #553, opened 2020-10-12
Fix libpng installation

seemethere:
We were missing zlib-devel headers as well as not CD'ing into the
correct directory

Follow up to #547 


andfoy (approved):
LGTM!

PR #559, opened 2020-10-23
Add script for generate sdist package for Windows platform in PyPI

gunandrose4u:
In this PR, provide a bash shell to generate a PyTorch downloading proxy for Windows platform. The generated torch-*.tar.gz package will be uploaded to PyPI, then Windows user can install torch from PyPI. Keep same install experience as user on Linux.
Eg.
bash ./generate_win_pypi_pkg.sh 1.6.0
pip install torch/dist/torch-1.6.0.tar.gz


packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+    cmdclass={
+        'bdist_wheel': bdist_wheel,
+        'install': install_torch,
+    },
+    packages=find_packages(),
+    url='https://pytorch.org/',
+    download_url='https://github.com/pytorch/pytorch/tags',
+    author='PyTorch Team',
+    author_email='packages@pytorch.org',
+    python_requires='>={},<{}'.format(python_min_version_str, python_max_version_str),
```
nbcsm, line 45: let's remove max version restriction, keep it consistent with torch.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+            f"Can not install torch due to local wheel built failed."
+            f"Plesse visit {torch_win_url} for more details."
+        )
+
+class bdist_wheel(wheel.bdist_wheel.bdist_wheel):
+    def run(self):
+        subprocess.check_call(
+            [sys.executable,'-m', 'pip', 'download', f'{pkg_name}==={pkg_ver}', '-f', torch_win_url, '--no-deps', '-d', self.dist_dir])
+
+setup(
```
nbcsm, line 31: need to specify `install_requires`, at least we need `wheel` to be there.
may also need to check min version required for wheel.

packaging/windows/build_pypi_pkg/setup.py
nbcsm, line 45: let's remove max version restriction, keep it consistent with torch.

nbcsm (changes requested):

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
nbcsm, line 1: could you add more information about the context to the PR description?
also, a sample command.

nbcsm (changes requested):

nbcsm:
@malfet , @peterjc123

malfet:
Since this wrapper is semantically not specific to Windows, but rather just a downloading proxy, please consider renaming the script

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+import subprocess
+
+import setuptools.command.install
+from setuptools import find_packages, setup
+import wheel.bdist_wheel
+
+pkg_name="torch"
+pkg_ver="{{GENERATE_TORCH_PKG_VER}}"
+torch_win_url="https://download.pytorch.org/whl/torch_stable.html"
+
```
nbcsm, line 11: nit: add space before/after =

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+        setuptools.command.install.install.run(self)
+        raise Exception(
+            f"Can not install torch due to local wheel built failed."
+            f"Plesse visit {torch_win_url} for more details."
+        )
+
+class bdist_wheel(wheel.bdist_wheel.bdist_wheel):
+    def run(self):
+        subprocess.check_call(
+            [sys.executable,'-m', 'pip', 'download', f'{pkg_name}==={pkg_ver}', '-f', torch_win_url, '--no-deps', '-d', self.dist_dir])
```
peterjc123, line 29: Why download here? Also we should add sth like `--binary-only`, right?

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
+import sys
+import subprocess
+
+import setuptools.command.install
+from setuptools import find_packages, setup
+import wheel.bdist_wheel
+
+pkg_name="torch"
+pkg_ver="{{GENERATE_TORCH_PKG_VER}}"
```
peterjc123, line 9: Do we need to generate the source package for every release of PyTorch?

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
```diff
@@ -0,0 +1,14 @@
...
+SETUP_FILE="setup.py"
+REPLACE_KEYWORD="{{GENERATE_TORCH_PKG_VER}}"
+
+mkdir torch
+cp $SETUP_FILE torch/
+cd torch
+sed -i "s/$REPLACE_KEYWORD/$TORCH_VER/g" $SETUP_FILE
+python $SETUP_FILE sdist
+
+echo "Generate package under torch/dist"
```
peterjc123, line 14: nit: EOL

peterjc123 (changes requested):
Thanks for the PR. I have some ideas regarding the changes you made.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+python_min_version_str = '.'.join((str(num) for num in python_min_version))
+python_max_version = (3, 9, 0)
+python_max_version_str = '.'.join((str(num) for num in python_max_version))
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        setuptools.command.install.install.run(self)
+        raise Exception(
+            f"Can not install torch due to local wheel built failed."
+            f"Plesse visit {torch_win_url} for more details."
```
peterjc123, line 23: typo: Plesse->Please

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Resolved

gunandrose4u:
> Since this wrapper is semantically not specific to Windows, but rather just a downloading proxy, please consider renaming the script

sure

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
gunandrose4u: Temp workaround for Windows before NV cudatoolkit is published to pypi, no aim to become general. These scripts will be removed after cudatoolkit published.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
+import sys
+import subprocess
+
+import setuptools.command.install
+from setuptools import find_packages, setup
+import wheel.bdist_wheel
+
+pkg_name="torch"
+pkg_ver="{{GENERATE_TORCH_PKG_VER}}"
+torch_win_url="https://download.pytorch.org/whl/torch_stable.html"
```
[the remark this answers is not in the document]
gunandrose4u: Temp workaround for Windows before NV cudatoolkit is published to pypi, no aim to become general. These scripts will be removed after cudatoolkit published.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+
+python_min_version = (3, 6, 1)
+python_min_version_str = '.'.join((str(num) for num in python_min_version))
+python_max_version = (3, 9, 0)
+python_max_version_str = '.'.join((str(num) for num in python_max_version))
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        setuptools.command.install.install.run(self)
+        raise Exception(
```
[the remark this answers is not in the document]
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+from setuptools import find_packages, setup
+import wheel.bdist_wheel
+
+pkg_name="torch"
+pkg_ver="{{GENERATE_TORCH_PKG_VER}}"
+torch_win_url="https://download.pytorch.org/whl/torch_stable.html"
+
+
+python_min_version = (3, 6, 1)
+python_min_version_str = '.'.join((str(num) for num in python_min_version))
```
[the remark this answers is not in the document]
gunandrose4u: This script is just for generate a tar.gz package for PT 1.7 release. So only 1.7 and 1.7+cpu gz package will be uploaded to PyPI.  So I follow up the setup.py under PyTorch to set python min_ver limitation.

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Resolved.

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: This one is just for PT1.7. It is a temp solution to work round PyPI package size limitation.

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123: > So only 1.7 and 1.7+cpu gz package will be uploaded to PyPI.

Could you please explain why do we need two packages? Users usually type in `pip install torch`, right?

peterjc123 (commented):
More comments coming...

packaging/windows/build_pypi_pkg/setup.py
nbcsm: we only need `wheel` here when installing our **fake** torch.
The **real** torch package already specified its **real** dependencies, so no need to include them here.

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
nbcsm: there will be only 1 package, the default CUDA one.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,80 @@
...
+    'wheel',
+    'numpy',
+    'future',
+    'typing_extensions',
+    'dataclasses; python_version < "3.7"'
+]
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        if sys.maxsize.bit_length() == 31:
```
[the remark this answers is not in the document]
gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
gunandrose4u: Yep, as @nbcsm said, will only for cuda one.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,80 @@
...
+            raise UserWarning("We don't support Python x86." \
+                "Please install Python x64 instead.")
+
+        if sys.platform == 'win32':
+            raise UserWarning("Please install using `python "\
+                "-m pip install xxxx -f https://download.pytorch.org/whl/torch_stable.html`.")
+
+        raise UserWarning(
+            f"Can not install torch due to local wheel built failed."
```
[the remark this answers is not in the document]
gunandrose4u: Resolved, change the error message here.
And other OS user, after they run pip install torch. They will not get tar.gz package generated by this file. So in this setup.py, we do not need handle other OS situation.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,80 @@
...
+            f"Please visit {torch_download_url} for more details."
+        )
+
+class bdist_wheel(wheel.bdist_wheel.bdist_wheel):
+    def run(self):
+        subprocess.check_call(
```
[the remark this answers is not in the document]
gunandrose4u: Same as above.

peterjc123:
@malfet Please take a look again.

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123: I didn't see any commit here. Did you forget to push?

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123: So you are trying to provide a Windows-only source package, aren't you?

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
gunandrose4u: Just pushed the commit :)

packaging/windows/build_pypi_pkg/setup.py
peterjc123: BTW, I still don't understand why do we call `pip download` here.

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123: That sounds reasonable.

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
gunandrose4u: Yes, just for Windows.

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -22,16 +22,16 @@
...
     def run(self):
-        if sys.maxsize.bit_length() == 31:
-            raise UserWarning("We don't support Python x86." \
-                "Please install Python x64 instead.")
-
         if sys.platform == 'win32':
             raise UserWarning("Please install using `python "\
                 "-m pip install xxxx -f https://download.pytorch.org/whl/torch_stable.html`.")
 
+        if sys.maxsize.bit_length() == 31:
```
peterjc123, line 29: Sorry, I misunderstood your previous comment. If it is a Windows-only package, then you shouldn't swap the order here. Determining x86 / x64 Python environment first is more user-friendly.

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
```diff
@@ -0,0 +1,14 @@
...
+
+TORCH_VER=$1
+SETUP_FILE="setup.py"
+REPLACE_KEYWORD="{{GENERATE_TORCH_PKG_VER}}"
+
+mkdir torch
+cp $SETUP_FILE torch/
+cd torch
+sed -i "s/$REPLACE_KEYWORD/$TORCH_VER/g" $SETUP_FILE
+python $SETUP_FILE sdist
```
peterjc123, line 12: If this is a Windows only package, maybe you should provide more args otherwise it will generate a source package for all platforms.

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: At first, we use pip install here directly, but it will introduce a issue that when user pip uninstall torch, files under caffe2 and torch will not be deleted. The reason is that in RECORD file, it will have file install records for this tar.gz package, without files installed under caffe2 and torch folder. So we changed to pip download, to make pip uninstall torch command fully work as on other platform.

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
peterjc123: More details about this is written in [PEP-425](https://www.python.org/dev/peps/pep-0425/#id14).

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,80 @@
...
+    'wheel',
+    'numpy',
+    'future',
+    'typing_extensions',
+    'dataclasses; python_version < "3.7"'
+]
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        if sys.platform == 'win32':
```
peterjc123, line 25: Do we need this if-condition?

gunandrose4u: Actually, does not need one. Since non-Windows platform will not touch this package.

packaging/windows/build_pypi_pkg/generate_win_pypi_pkg.sh
peterjc123: I did a google search and was amazed to find out that it is currently impossible to generate a platform-specific sdist package. See https://www.python.org/dev/peps/pep-0625/ and https://docs.python.org/3/distutils/sourcedist.html .

gunandrose4u: Yep, I just want to find sdist options, does not find one.

packaging/windows/build_pypi_pkg/setup.py
gunandrose4u: Will remove the sys.platform condition since we do not need it. So only 64bit python detection will happen.

peterjc123 (commented):
The only one and the most important question remaining:
**How to make this package Windows only?**

nbcsm:
> The only one and the most important question remaining:
> **How to make this package Windows only?**

pip will handle this automatically.
since there are already wheels for Linux and MacOS, but not wheel for Windows.
so when user calls `pip install torch` on Windows, this source distribution (.gz) will be selected.
when user calls `pip install torch` on Linux or MacOS, corresponding wheel (.whl) will be selected.

if you still have concern, we can throw error in this source distribution when current OS is NOT windows. :-)

peterjc123:
> if you still have concern, we can throw error in this source distribution when current OS is NOT windows. :-)

I'm okay to that. Then the code should look like sth like below.
```py
if sys.platform == 'win32':
   if not is_64bit:
        ....

raise UserWarning("Cannot find a binary package for your Python environment.")
```

Also, you need to protect `bdist_wheel` with a Windows-only if-condition.

gunandrose4u:
> > if you still have concern, we can throw error in this source distribution when current OS is NOT windows. :-)
> 
> I'm okay with that. Then the code should look like sth like below.
> 
> ```python
> if sys.platform == 'win32':
>    if not is_64bit:
>         ....
> 
> raise UserWarning("Cannot find a binary package of PyTorch for your Python environment.")
> ```
> 
> Also, you need to protect `bdist_wheel` with a Windows-only if-condition, otherwise revert to default behavior.

I added option '--platform', 'win_amd64' in pip download command to enforce only PyTorch wheel for 64-bit Windows will be downloaded.

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123:
```suggestion
        if sys.platform == 'win32' and sys.maxsize.bit_length() == 31:
```

peterjc123:
@seemethere @malfet @soumith Would you please review this PR?

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
peterjc123: @gunandrose4u Please accept the suggested change.

gunandrose4u:
> Since this wrapper is semantically not specific to Windows, but rather just a downloading proxy, please consider renaming the script

Resolved, rename the folder and bash file with a download proxy. Since this one is just for generate a PT Windows package downloading proxy, so still keep Windows in file name.

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -0,0 +1,79 @@
...
+    'wheel',
+    'numpy',
+    'future',
+    'typing_extensions',
+    'dataclasses; python_version < "3.7"'
+]
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        if sys.maxsize.bit_length() == 31:
```
peterjc123, line 25: Please make this if-condition Windows-only.

gunandrose4u: Resolved

packaging/windows/build_pypi_pkg/setup.py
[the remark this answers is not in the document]
gunandrose4u: Resolved

peterjc123 (approved):

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -0,0 +1,79 @@
...
+
+python_min_version = (3, 6, 1)
+python_min_version_str = '.'.join((str(num) for num in python_min_version))
+
+install_requires = [
+    'wheel',
+    'numpy',
+    'future',
+    'typing_extensions',
+    'dataclasses; python_version < "3.7"'
```
malfet, line 20: Why should the wrapper package depend on anything but pip? (and 'pip' depends on 'wheel')
```suggestion
   'pip',
```

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -0,0 +1,79 @@
...
+]
+
+class install_torch(setuptools.command.install.install):
+    def run(self):
+        if sys.platform == 'win32' and sys.maxsize.bit_length() == 31:
```
malfet, line 25: Why limit this check to win32 rather than to `platform.machine() in ['i386','x86_64']` (as error message suggests)

malfet (approved):

packaging/windows/build_pypi_pkg_download_proxy/setup.py
malfet: Hmm, why should it be windows specific? Pytorch is not supported on all x86 32-bit platforms

packaging/windows/build_pypi_pkg/setup.py
```diff
@@ -0,0 +1,64 @@
...
+import wheel.bdist_wheel
+
+pkg_name="torch"
+pkg_ver="{{GENERATE_TORCH_PKG_VER}}"
+torch_win_url="https://download.pytorch.org/whl/torch_stable.html"
+
+
+python_min_version = (3, 6, 1)
+python_min_version_str = '.'.join((str(num) for num in python_min_version))
+python_max_version = (3, 9, 0)
```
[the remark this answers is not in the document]
gunandrose4u: Looks like github web does sync latest code. I've removed python_max_version in commit https://github.com/pytorch/builder/pull/559/commits/5494154fad474b1e7d705f4dc31532fabe41b987.

packaging/windows/build_pypi_pkg_download_proxy/setup.py
gunandrose4u: Here is to restrict generated package can only run on Windows platform (64bit). Maybe we can refine the error message to make it more specific. How do you think?

PR #561, opened 2020-10-29
add ROCm 3.9 manywheel build

jeffdaily:


jeffdaily:
@walterddr could you please review?  Thanks.

manywheel/build_rocm.sh
```diff
@@ -59,22 +59,31 @@ elif [[ "$OS_NAME" == *"Ubuntu"* ]]; then
     LIBELF_PATH="/usr/lib/x86_64-linux-gnu/libelf.so.1"
 fi
 
+# rocm3.8 and later use TensileLibrary.dat
 if [[ $ROCM_VERSION == "rocm3.7" ]]; then
     TENSILE_LIBRARY_NAME=TensileLibrary.yaml
-elif [[ $ROCM_VERSION == "rocm3.8" ]]; then
+else
```
walterddr, line 65:
```suggestion
elif [[ $ROCM_VERSION == "rocm3.8" || $ROCM_VERSION == "rocm3.9" ]]; then
```

jeffdaily: I did consider that, but IIUC the tensile library name extension will hereafter always be *.dat, so the rocm3.7 case was the only exceptional case.

walterddr: got it. thanks for the explanation

walterddr (approved):

walterddr:
@jeffdaily did you have rocm3.9 docker images created?

Issue #530, opened 2020-09-24
[conda] Should ninja be a runtime dependency?


alanhdu:
Right now, if you `conda install -c pytorch pytorch`, you also pull in `ninja` as a run-time dependency. I don't think this should be the case though -- Ninja is a build tool, so shouldn't it just be a build-time dependency?

I believe we can fix this by removing `ninja` from the run dependencies:

https://github.com/pytorch/builder/blob/bb24ca1110ce2d09700bfa29b8840535cb7d66e9/conda/pytorch-nightly/meta.yaml#L27-L31

scopatz:
Yeah, I agree. Ninja should not be a runtime dep

PR #562, opened 2020-10-30
Remove ninja as a runtime conda dependency

alanhdu:
Closes https://github.com/pytorch/builder/issues/530


alanhdu:
I believe that the CI failures are unrelated to this change.

rgommers (approved):
This change looks correct to me. `ninja` may be needed for `cpp_extension`, but should be an optional dependency (just like `numpy` is now) and not installed by default. 

The one thing to check is if `ninja` is explicitly installed in CI if this package is used anywhere in pytorch CI.

alanhdu:
Rebased to remove the merge conflict.

malfet (approved):

PR #566, opened 2020-11-09
Determine min Python version based on PyTorch version

skyline75489:
For PyTorch version > 1.7, the min Python version should be `3.6.2`.  This aligns with the `setup.py` in current master https://github.com/pytorch/pytorch/blob/9e0102c10f57631000ebf28e55a055f480b9b780/setup.py#L179 .

If you're wondering what the meaning differences between `3.6.1` and `3.6.2` are, one of them would be [typing.NoReturn](https://docs.python.org/3/library/typing.html#typing.NoReturn), which was first added in Python 3.6.2. This actually causes build error if you are using `3.6.1` and try to install from source on current master.

### DIFF
--- a/packaging/windows/build_pypi_pkg_download_proxy/setup.py
+++ b/packaging/windows/build_pypi_pkg_download_proxy/setup.py
@@ -3,6 +3,7 @@ import subprocess
 
 import setuptools.command.install
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


skyline75489:
Close this due to the fact we are mainly targeting 1.7.0. By the time 1.8.0 is out the cudatoolkit should already make its way into PyPI.

PR #568, opened 2020-11-10
Fix setup.py for pypi sdist package on Windows

skyline75489:
The original code in `install_torch` is broken. It will raise an error regardless of the platform, even on win32-amd64 which we are targeting. 

This PR fixes that.


skyline75489:
CC @gunandrose4u  @nbcsm  @peterjc123

peterjc123 (approved):

peterjc123:
@malfet @soumith @seemethere Please take a look.

skyline75489:
I was experimenting with this PR for a while. Here's what I have learned so far. 

### Baseline

The elephant in the room, is that `pip install torch==1.7.0` fails on Windows, with this error message:

```plaintext
ERROR: Could not find a version that satisfies the requirement torch==1.7.0 (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2)
ERROR: No matching distribution found for torch==1.7.0
```

### Solution

We proposed #559 as a possible solution. Together with this PR, the `setup.py` will create a working "redirection" package for users to consume.

### Testing

I build the `sdist` using the script in #559 , targeting version `1.7.0`.

Due to the fact there's already a package named `torch` in TestPyPI, I choose to host my own PyPi server the simple way, that is to build a static content server which can be consumed by `pip`. The manual can be found [here](https://packaging.python.org/guides/hosting-your-own-index/#hosting-your-own-simple-repository).

To mimic the behaviour of official PyPI, I've uploaded the generated `sdist` to the server, put it under the `torch` directory alongside with the existing wheels I downloaded from official PyPI. The directory structure looks like this:

```plaintext
torch/
  torch-1.7.0-cp36-cp36m-manylinux1_x86_64.whl | 740M |  
  torch-1.7.0-cp36-none-macosx_10_9_x86_64.whl | 103M |  
  torch-1.7.0-cp37-cp37m-manylinux1_x86_64.whl | 740M |  
  torch-1.7.0-cp37-none-macosx_10_9_x86_64.whl | 103M |  
  torch-1.7.0-cp38-cp38-manylinux1_x86_64.whl | 740M | 
  torch-1.7.0-cp38-none-macosx_10_9_x86_64.whl | 103M | 
  torch-1.7.0.tar.gz | 1K |
```

I've also put required dependencies such as NumPy on my server to make it more "official".

I've tested several cases under different environment. I use the following command to install the package:

```plaintext
>pip install -i http://myserver.com:8080 --trusted-host myserver.com torch==1.7.0
```

Here's what I got so far:
 
✔ means installation is a success.

|  | 3.6 |3.7 | 3.8|
| --|--|--|--|
|Win| ✔|✔|✔|
|Linux| ✔|✔|✔|
|macOS| ✔|✔|✔|

It's boringly good. The installation experience on Windows is basically the same as on other platforms.

### Caveats

There is one issue that caused me headache, is that due to the nature of a simple static server, my server is not able to provide metadata about the package for [dataclasses](https://pypi.org/pypi/dataclasses/json). This leads me to https://github.com/pytorch/pytorch/issues/46930 . Due to the lack of dependency requirement in metadata,  `pip` will always install the latest version of dataclasses(0.7) that simply won't work on Python3.7+. I had to install dataclasses from the official PyPI to get the correct version(0.6) , then proceed with my experiments.

This should not be a problem once the `sdist` is in official PyPI.

### Concerns

* Does uninstall / reinstall work? 
  
  To my knowledge it works as expected.

* If the installation is interrupted by `Ctrl+C`, will it still work?

  To my knowledge it works. The subprocess (`pip download`) will also be terminated. Reinstall it and it will restart the download.

* Does CUDA work?

  Gladly yes, as long as the platform has the required device.

peterjc123:
> I was experimenting with this PR for a while. Here's what I have learned so far.
> 
> ### Baseline
> The elephant in the room, is that `pip install torch==1.7.0` fails on Windows, with this error message:
> 
> ```
> ERROR: Could not find a version that satisfies the requirement torch==1.7.0 (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2)
> ERROR: No matching distribution found for torch==1.7.0
> ```
> 
> ### Solution
> We proposed #559 as a possible solution. Together with this PR, the `setup.py` will create a working "redirection" package for users to consume.
> 
> ### Testing
> I build the `sdist` using the script in #559 , targeting version `1.7.0`.
> 
> Due to the fact there's already a package named `torch` in TestPyPI, I choose to host my own PyPi server the simple way, that is to build a static content server which can be consumed by `pip`. The manual can be found [here](https://packaging.python.org/guides/hosting-your-own-index/#hosting-your-own-simple-repository).
> 
> To mimic the behaviour of official PyPI, I've uploaded the generated `sdist` to the server, put it under the `torch` directory alongside with the existing wheels I downloaded from official PyPI. The directory structure looks like this:
> 
> ```
> torch/
>   torch-1.7.0-cp36-cp36m-manylinux1_x86_64.whl | 740M |  
>   torch-1.7.0-cp36-none-macosx_10_9_x86_64.whl | 103M |  
>   torch-1.7.0-cp37-cp37m-manylinux1_x86_64.whl | 740M |  
>   torch-1.7.0-cp37-none-macosx_10_9_x86_64.whl | 103M |  
>   torch-1.7.0-cp38-cp38-manylinux1_x86_64.whl | 740M | 
>   torch-1.7.0-cp38-none-macosx_10_9_x86_64.whl | 103M | 
>   torch-1.7.0.tar.gz | 1K |
> ```
> 
> I've also put required dependencies such as NumPy on my server to make it more "official".
> 
> I've tested several cases under different environment. I use the following command to install the package:
> 
> ```
> >pip install -i http://myserver.com:8080 --trusted-host myserver.com torch==1.7.0
> ```
> 
> Here's what I got so far:
> 
> ✔ means installation is a success.
> 
> 3.6	3.7	3.8
> Win	✔	✔	✔
> Linux	✔	✔	✔
> macOS	✔	✔	✔
> It's boringly good. The installation experience on Windows is basically the same as on other platforms.
> 
> ### Caveats
> There is one issue that caused me headache, is that due to the nature of a simple static server, my server is not able to provide metadata about the package for [dataclasses](https://pypi.org/pypi/dataclasses/json). This leads me to [pytorch/pytorch#46930](https://github.com/pytorch/pytorch/issues/46930) . Due to the lack of dependency requirement in metadata, `pip` will always install the latest version of dataclasses(0.7) that simply won't work on Python3.7+. I had to install dataclasses from the official PyPI to get the correct version(0.6) , then proceed with my experiments.
> 
> This should not be a problem once the `sdist` is in official PyPI.
> 
> ### Concerns
> * Does uninstall / reinstall work?
>   To my knowledge it works as expected.
> * If the installation is interrupted by `Ctrl+C`, will it still work?
>   To my knowledge it works. The subprocess (`pip download`) will also be terminated. Reinstall it and it will restart the download.
> * Does CUDA work?
>   Gladly yes, as long as the platform has the required device.

Thanks for the effort into testing the package. The results look promising. Yet, I still have some questions to ask.
1. Does it work when written in `requirements.txt` and `pip install -r requirements.txt`?
2. Will it break the support for legacy Python versions? For example, I'm a Python 3.5 user and and say that I want to install the latest supported PyTorch (1.6.0). And then he type in sth like `pip install torch` or `pip install torch>=1.6.0`. Is this case still working?
3. Which package will be downloaded for Windows? The one with CUDA 10.2, right?
4. Does upgrading / downgrading from/to the package work?

Thanks again for your report. However, uploding a sdist package has a potential to affect the user experience on all the operation systems, so we need to be very careful.

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -14,18 +14,23 @@
 
 install_requires = [
    'pip',
+   'future',
```
malfet, line 17: `future` shouldn't be needed since we are only targeting Python-3.6+, aren't we?
```suggestion
```

malfet (approved):

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -14,18 +14,23 @@
...
+   'numpy',
+   'typing_extensions',
+   'dataclasses; python_version < "3.7"'
```
malfet, line 20: Why any of those packages are needed for the proxy?

skyline75489: The wheel is downloaded with no-deps flag. On a clean Python environment, without these dependencies, the installation would fail

packaging/windows/build_pypi_pkg_download_proxy/setup.py
skyline75489: This unfortunately is required in the wheel so I had to keep it here.

skyline75489:
>Does it work when written in requirements.txt and pip install -r requirements.txt?

Yes, it works with `requirements.txt`. I had to add my server to it, though:

```plaintext
--index-url http://52.234.131.30:8080 
torch==1.7.0
```

And install it using `pip install -U --trusted-host 52.234.131.30 -r requirement.txt`.

>Will it break the support for legacy Python versions? For example, I'm a Python 3.5 user and and say that I want to install the latest supported PyTorch (1.6.0). And then he type in sth like pip install torch or pip install torch>=1.6.0. Is this case still working?

Per the conversation in #559 I think that we only need to support 1.7.0, which requires Python 3.6+. In this case `pip` should correctly recognize that 1.7.0 is not compatible with 3.5.

>Which package will be downloaded for Windows? The one with CUDA 10.2, right?

Yes. The exact file is `cu102/torch-1.7.0-cp37-cp37m-win_amd64.whl`.

>Does upgrading / downgrading from/to the current version work?

I built another sdist with 1.6.0 and try the upgrading & downgrading from/to 1.7.0, it works as expected.

---

>uploding a sdist package has a potential to affect the user experience on all the operation systems.

I understand the concern. As far as I can tell, due to fact there are existing wheels for Linux & macOS, the newly added `torch-1.7.0.tar.gz` sdist only affects Windows users. It seems that `pip` will always choose the wheel that matches the current platform over the sdist. You can also tell this from the original error message:

>ERROR: Could not find a version that satisfies the requirement torch==1.7.0 (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2)

These available versions on Windows are exactly the only three sdist you can find in the official [torch PyPI index](https://pypi.org/simple/torch/):

```plaintext
Links for torch

torch-0.1.2.tar.gz
torch-0.1.2.post1.tar.gz
torch-0.1.2.post2.tar.gz
torch-0.3.0.post4-cp27-none-macosx_10_6_x86_64.whl
...
(all the other wheels)
```

I think they have been there, well, since version 0.1.2. And everyone seems to be OK with it.

In conclusion, adding another `torch-1.7.0.tar.gz` as we proposed in this series of PRs should have almost no impact on users on other platforms.

peterjc123:
> Per the conversation in #559 I think that we only need to support 1.7.0, which requires Python 3.6+. In this case `pip` should correctly recognize that 1.7.0 is not compatible with 3.5.

@malfet Could you please confirm this is the case? I want to make sure because in my understanding, this is broken. `pip` will pick a package according to the version first and then it will prefer binaries to sources. Since 1.7.0 > 1.6.0, so users will get this package instead.

peterjc123:
> understand the concern. As far as I can tell, due to fact there are existing wheels for Linux & macOS, the newly added `torch-1.7.0.tar.gz` sdist only affects Windows users.

The binary packages has a platform specific tag. (e.g. `-win-amd64`) So they won't affect the users on other OSes in any case. That's why I say it is safe.

skyline75489:
@peterjc123 I've tested it on my simple PyPI server with Python 3.5. This is what I got:

```plaintext
>pip install -U -i http://myserver.com8080 --trusted-host myserver.com torch==1.7.0
Looking in indexes: http://myserver.com:8080
Collecting torch==1.7.0
  Downloading http://myserver.com:8080/torch/torch-1.7.0.tar.gz
torch requires Python '>=3.6.1' but the running Python is 3.5.6
```

I had to change all the f-strings expression to old-style format because it was added Python 3.6. Otherwise `setup.py` will throw syntax error:

```plaintext
Looking in indexes: http://myserver.com:8080
Collecting torch==1.7.0
  Downloading http://myserver.com:8080/torch/torch-1.7.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\jialli\AppData\Local\Temp\pip-install-zpt5pu35\torch\setup.py", line 30
        f"Can not download torch binary from {torch_download_url}." \
                                                                  ^
    SyntaxError: invalid syntax

```

peterjc123:
> @peterjc123 I've tested it on my simple PyPI server with Python 3.5. This is what I got:
> 
> ```
> >pip install -U -i http://myserver.com8080 --trusted-host myserver.com torch==1.7.0
> Looking in indexes: http://myserver.com:8080
> Collecting torch==1.7.0
>   Downloading http://myserver.com:8080/torch/torch-1.7.0.tar.gz
> torch requires Python '>=3.6.1' but the running Python is 3.5.6
> ```
> 
> I had to change all the f-strings expression to old-style format because it was added Python 3.6. Otherwise `setup.py` will throw syntax error:
> 
> ```
> Looking in indexes: http://myserver.com:8080
> Collecting torch==1.7.0
>   Downloading http://myserver.com:8080/torch/torch-1.7.0.tar.gz
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "C:\Users\jialli\AppData\Local\Temp\pip-install-zpt5pu35\torch\setup.py", line 30
>         f"Can not download torch binary from {torch_download_url}." \
>                                                                   ^
>     SyntaxError: invalid syntax
> ```

Yeah, it meets my expectation. I don't think it will a big block here. Since the easy fix is to publish an addition 1.6.0 sdist package for Python 3.5 users.

skyline75489:
Should I change the f-strings, though? I'm personally OK with changing it.

packaging/windows/build_pypi_pkg_download_proxy/setup.py
```diff
@@ -14,18 +14,23 @@
...
-            f"Please visit {torch_download_url} for more details."
-        )
+        if sys.platform == 'win32':
+            if sys.maxsize.bit_length() == 31:
+                raise UserWarning("We don't support Python x86." \
+                    "Please install Python x64 instead.")
+        else:
+            raise UserWarning(
+                f"Can not download torch binary from {torch_download_url}." \
+                f"Please visit {torch_download_url} for more details."
```
seemethere, line 32: Perhaps we should point them towards https://pytorch.org/get-started/locally/ instead so they can grab the installation instructions from there

packaging/windows/build_pypi_pkg_download_proxy/setup.py
seemethere: Would it be possible to download with dependencies? Having extra dependencies to maintain here seems like it'd be a headache later on down the line.

packaging/windows/build_pypi_pkg_download_proxy/setup.py
skyline75489: Good idea!

packaging/windows/build_pypi_pkg_download_proxy/setup.py
skyline75489: As far as I can see, we are only doing this for the 1.7.x release. By the time 1.8.0 is out, the cudatoolkit should be ready on Windows and we won’t need this proxy package anymore. @nbcsm Can you confirm this?

skyline75489:
@seemethere About the concern of having extra dependencies, I've conducted more experiments with the script. Here's what I learned. 

Normally, when someone tries to install a `wheel` locally, for example using:

```plaintext
pip install torch-1.7.0-cp37-cp37m-manylinux1_x86_64.whl
```

Then `pip` will correctly install the dependencies guided by the `METADATA` part in the wheel itself. 

However, due to the fact we're using a somewhat hacky way (using a custom `bdist_wheel` command) to `pip download` the actual wheel, `pip` will not analyze or install the dependencies in the wheel anymore. I haven't figured out the underlying reason of this but I don't think there's a way to easily bypass this.

I've tried removing the `no-deps`, then `pip` would download extra wheels that do not match the name of the package "torch", then refuse to install it.

I've tried adding `download.pytorch.org` in `setup.py` as a extra index-url. The closest thing I could found is [dependency_links](https://stackoverflow.com/a/30064248). But it was [deprecated](https://stackoverflow.com/a/50591759) for [security reasons](https://mail.python.org/pipermail/distutils-sig/2013-October/022937.html).

I've tried using the PEP 508 style URL and shove it into `install_requires`, it does not work either.

I think putting the dependencies inside `setup.py` might be the only way to accomplish what we want.

PR #575, opened 2020-11-12
wheel: Create clean conda environment for testing

seemethere:
There was an issue where libiomp.dylib was specified in multiple
packages within the conda environment (with `llvm-openmp` being added as a dependency for libopenblas, which is a dependency of `numpy>=19` so let's just create a brand new
conda environment for testing so we can avoid all of these issues.

This should resolve issues of binaries built with Python 3.9 on macOS

cc @peterjc123 the same will have to be done with windows, but my batch scripting skills seems to be failing me right now.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -76,7 +76,7 @@ if [[ "$package_type" == conda || "$(uname)" == Darwin ]]; then
         fi
     fi
     if [[ "$(python --version 2>&1)" == *3.9.* ]]; then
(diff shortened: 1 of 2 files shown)


malfet (approved):

walterddr (approved):

xuzhao9 (approved):

peterjc123:
@seemethere Will do.

PR #581, opened 2020-11-16
Use simplest L1Loss() instead

walterddr:
Use simplest L1Loss() to smoke test backward step


walterddr:
still doesn't work. my bad

PR #587, opened 2020-11-17
Add sm_86 arch to CUDA 11.0 builds

seemethere:
Enables `sm_86` on CUDA 11.0 builds since we are now building with CUDNN `v8.0.5` which should include better support for `sm_86`

change is being tested here: https://github.com/pytorch/pytorch/pull/48128

### DIFF
--- a/manywheel/build.sh
+++ b/manywheel/build.sh
@@ -53,7 +53,7 @@ case ${CUDA_VERSION} in
         EXTRA_CAFFE2_CMAKE_FLAGS+=("-DATEN_NO_TEST=ON")
         ;;
     11.0)
-        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;7.5;8.0"
+        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;7.5;8.0;8.6"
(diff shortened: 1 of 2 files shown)


manywheel/build.sh
```diff
@@ -53,7 +53,7 @@ case ${CUDA_VERSION} in
         EXTRA_CAFFE2_CMAKE_FLAGS+=("-DATEN_NO_TEST=ON")
         ;;
     11.0)
-        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;7.5;8.0"
+        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;7.5;8.0;8.6"
```
ptrblck, line 56: This would try to enable `sm86` on CUDA11.0 and should thus fail. 
The previous 11.1 block already has 8.6 in the architecture list.

seemethere: Correct, the question I have though is if CUDA 11.1 is actually required for `sm_86` or is it just that CUDNN 8.0.5 is required?

ptrblck: CUDA11.1 is required. if you want to build for `sm_86`.

seemethere: Okay then, makes sense

PR #590, opened 2020-11-19
Use DESIRED_CUDA var in magma-cuda*/build.sh

janeyx99:
This is a first step in consolidating our MAGMA directory as there is a lot of repetition across the folders there. Using the DESIRED_CUDA environment variable would be a step in the direction of having one directory for magma instead of however many versions of CUDA we are supporting. 

I am less sure about how to consolidate the `.patch` and `.yaml` files and would love any pointers on that. 

Getting tested in https://github.com/pytorch/pytorch/pull/48259

### DIFF
--- a/magma/magma-cuda101/build.sh
+++ b/magma/magma-cuda101/build.sh
@@ -2,9 +2,9 @@ export CMAKE_LIBRARY_PATH=$PREFIX/lib:$PREFIX/include:$CMAKE_LIBRARY_PATH
 export CMAKE_PREFIX_PATH=$PREFIX
 export PATH=$PREFIX/bin:$PATH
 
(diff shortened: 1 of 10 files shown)


seemethere:
This might work, but we'd also need some way to propagate the DESIRED_CUDA environment variable to the conda build environment.

This might give some insight into that: https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#inherited-environment-variables

janeyx99:
Ah, should I also add the other environment variables that are used in build.sh to each meta.yaml? For example, `PREFIX` and `CMAKE_LIBRARY_PATH`, or are those defined within conda already?

seemethere:
Those should be defined already within the conda environment, but I don't believe that DESIRED_CUDA is already defined

seemethere (approved):

PR #591, opened 2020-11-19
adding sm_86 to magma for cuda11.1 in cmakelists.patch

janeyx99:
CUDA 11.1 can now support sm_86.


seemethere:
Does this patch still apply cleanly? I noticed you had removed a lot of whitespace.

janeyx99:
My editor removes trailing whitespaces on save--does that mess with patches even though the new lines are preserved?

seemethere:
> My editor removes trailing whitespaces on save--does that mess with patches even though the new lines are preserved?

It might, you should try to build it at least once to verify that it still functions as expected

janeyx99:
Successfully ran locally and manually checked that the output looked the same (same files in subdirectories).

seemethere (approved):

PR #595, opened 2020-11-20
Refactoring MAGMA build process

janeyx99:
This PR attempts to refactor the much duplicated magma build process.

1. It moves the process to a Makefile for better organization and passing of environment variables
2. It removes the duplicated folders and generates a folder to package when run
3. It uses environment variables to consolidate cmakelists.patch and meta.yml
4. It assumes that all magma-cuda versions do not need magma sparse as PyTorch does not use magma sparse.


magma/package_files/cmakelists.patch
```diff
@@ -58,7 +58,7 @@
 
 +    if ( ${GPU_TARGET} MATCHES "All")
 +      set( MIN_ARCH 370)
-+      SET( NV_SM -gencode arch=compute_37,code=sm_37 -gencode arch=compute_50,code=sm_50 -gencode arch=compute_60,code=sm_60 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_80,code=sm_80 -gencode arch=compute_86,code=sm_86)
++      SET( NV_SM ${CUDA_ARCH_LIST})
```
samestep, line 61: I don't entirely understand how `cmakelists.patch` works here, but is it relevant that there seems to be a pretty big difference between `magma-cuda102/cmakelists.patch` and `magma-cuda110/cmakelists.patch` (in base), not just this one line? Specifically, when I compare those files locally, I see the following:
- `magma-cuda92/cmakelists.patch`, `magma-cuda101/cmakelists.patch`, and `magma-cuda102/cmakelists.patch` are all identical
- `magma-cuda102/cmakelists.patch` differs from `magma-cuda110/cmakelists.patch` in its `CUDA_ARCH_LIST`, but also in several other ways
- `magma-cuda110/cmakelists.patch` and `magma-cuda111/cmakelists.patch` are identical except their `CUDA_ARCH_LIST`s and some trailing whitespace in the former

I guess my question is, why do you need a `CUDA_ARCH_LIST` variable to account for this specific difference among the `cmakelists.patch` files, but nothing to account for the other differences between `magma-cuda102/cmakelists.patch` and `magma-cuda110/cmakelists.patch`?

seemethere (approved):

samestep (approved):

magma/package_files/cmakelists.patch
janeyx99: Ah good question. The rest of the cmakelists.patch edits referred to magma sparse. Since we don't use magma sparse, it was safe to remove those for the older CUDA versions too.

PR #599, opened 2020-11-30
refine build script

walterddr:
mainly refine key-name setting and add exceptions for AMI not found

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -38,7 +38,7 @@ def ec2_instances_by_id(instance_id):
     return rc[0] if len(rc) > 0 else None
 
 
-def start_instance(ami=ubuntu18_04_ami, instance_type='t4g.2xlarge', key_name=None):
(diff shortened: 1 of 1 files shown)


build_aarch64_wheel.py
```diff
@@ -36,7 +38,7 @@ def ec2_instances_by_id(instance_id):
     return rc[0] if len(rc) > 0 else None
 
 
-def start_instance(ami=ubuntu18_04_ami, instance_type='t4g.2xlarge', key_name='nshulga-key'):
+def start_instance(ami=ubuntu18_04_ami, instance_type='t4g.2xlarge', key_name=None):
```
malfet, line 41: Hmm, defaulting `key_name` to `None` seems wrong...

malfet (approved):

build_aarch64_wheel.py
walterddr: we can remove the default and move the argument to the front?

PR #601, opened 2020-12-02
Remove run_test.sh from MAGMA files as it's a file of commented out code

janeyx99:
run_test.sh is optional and unused currently


samestep (approved):
lol

PR #603, opened 2020-12-03
add rocm 3.10 manywheel support

jeffdaily:


jeffdaily:
@walterddr if you would review, thanks.

walterddr (approved):

PR #607, opened 2020-12-08
s3_management: Add tools to generate HTML indices for S3

seemethere:
Refactors our S3 HTML index generation script to be written in python and have extra functionality like limiting nightly builds to a specific threshold


s3_management/manage.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+S3 = boto3.resource('s3')
+CLIENT = boto3.client('s3')
+BUCKET = S3.Bucket('pytorch')
+
+ACCEPTED_FILE_EXTENSIONS = ("whl", "zip")
+
+# How many packages should we keep of a specific package?
+KEEP_THRESHOLD = 20
```
malfet, line 19: Can we have a threshold of say 60 days?

malfet (approved):
LGTM

s3_management/prune_nightly.sh
```diff
@@ -0,0 +1,21 @@
...
+if [[ ${DRY_RUN} = "disabled" ]]; then
+  DRY_RUN_FLAG=""
+fi
+
+for pkg in $(cat packages_to_delete); do
+  (
+    set -x
+    aws s3 rm ${DRY_RUN_FLAG} "${pkg}"
+  )
+done
```
malfet, line 21: Hmm, why not delete it right from the .py script?

seemethere: I wanted to have an option to do a dry run for testing purposes and it turns out that boto3 does not have a dry run option.

seemethere: see: https://stackoverflow.com/questions/57040302/how-to-run-boto3-run-job-flow-in-a-dry-run

PR #610, opened 2020-12-10
Add rules for producing release builds

malfet:
Also, add python-3.9 as an option

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -161,27 +161,35 @@ def start_build(key_name, *,
     print('Checking out PyTorch repo')
     run_ssh(addr, f"git clone --recurse-submodules -b {branch} https://github.com/pytorch/pytorch {git_clone_flags}")
     print('Building PyTorch wheel')
+    build_vars=""
     if branch == 'nightly':
         build_date = check_output(addr, "cd pytorch ; git log --pretty=format:%s -1").strip().split()[0].replace("-", "")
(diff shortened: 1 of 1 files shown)


build_aarch64_wheel.py
```diff
@@ -161,27 +161,35 @@ def start_build(key_name, *,
...
+    run_ssh(addr, f"cd pytorch ; {build_vars} python3 setup.py bdist_wheel")
     pytorch_wheel_name = list_dir(addr, "pytorch/dist")[0]
     embed_libgomp(addr, use_conda, os.path.join('pytorch', 'dist', pytorch_wheel_name))
     print('Copying the wheel')
     subprocess.check_call(["scp", "-i", keyfile_path, f"ubuntu@{addr}:pytorch/dist/*.whl", "."])
 
     print('Checking out TorchVision repo')
-    run_ssh(addr, f"git clone https://github.com/pytorch/vision {git_clone_flags}")
+    if branch.startswith("v1.7.1"):
+        run_ssh(addr, f"git clone https://github.com/pytorch/vision -b v0.8.2-rc2 {git_clone_flags}")
```
walterddr, line 179: interesting :-)

walterddr (approved):

seemethere (approved):

build_aarch64_wheel.py
malfet: I can't think of a better way of doing that, perhaps we can have a version correspondence dictionary later

PR #622, opened 2021-01-19
manywheel: Upgrade git for use with GHA

seemethere:
GHA requires a newer version of git to be installed in order to be able
to use the "submodule" functionality from "actions/checkout"

Example: https://github.com/pytorch/pytorch/pull/50633/checks?check_run_id=1730427448

unblocks https://github.com/pytorch/pytorch/pull/50633


seemethere:
cc @jeffdaily since the ROCM images will also need to be re-built after this goes through

PR #628, opened 2021-01-28
add support for ROCm 4.0.1, including custom MIOpen package

jeffdaily:
For ROCm versions 4.0.1 or greater, MIOpen is built from source with the Code Object Manager (COMgr) option enabled (default miopen-hip package has this disabled.)  This allows MIOpen to compiling kernels at runtime using COMgr instead of depending on a system call to hipcc or other installed compiler binaries.  For wheels, this drop-in MIOpen replacement becomes the bundled library.

### DIFF
--- a/common/install_miopen.sh
+++ b/common/install_miopen.sh
@@ -1,0 +1,115 @@
+#!/bin/bash
+
+set -ex
+
+ROCM_VERSION=$1
+
+if [[ -z $ROCM_VERSION ]]; then
+    echo "missing ROCM_VERSION"
+    exit 1;
+fi
+
+# To make version comparison easier, create an integer representation.
+save_IFS="$IFS"
+IFS=. ROCM_VERSION_ARRAY=(${ROCM_VERSION})
+IFS="$save_IFS"
+if [[ ${#ROCM_VERSION_ARRAY[@]} == 2 ]]; then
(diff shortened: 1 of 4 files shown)


jeffdaily:
This PR also revises the many ROCm version if/else condition blocks, simplifying them by using integer comparison.  As far as major *.so versions are concerned, very little changes from one release to the next, so much of the logic can be carefully reused from one version to the next.

jeffdaily:
FYI: new docker image for ROCm 4.01 has already been pushed.  `pytorch/manylinux-rocm:4.0.1`

malfet:
LGTM, but from a quick glance at `manywheel/Dockerfile` it looks like we always install both CUDA and RoCM into the same docker image. If this is the case, can we please add logic to install either one or the other? (In order to reduce image space)

jeffdaily:
The `manywheel/Dockerfile` uses docker buildkit (https://docs.docker.com/develop/develop-images/build_enhancements/).  CUDA and ROCm will be in the same Dockerfile, sharing what common code makes sense, but will produce distinct images.  See the `manywheel/deploy.sh` for details.

seemethere (approved):
We should also probably take an opportunity to remove some of the older rocm versions we don't build for

PR #632, opened 2021-02-01
conda: Set cudatoolkit 11.0 dependency minimum

seemethere:
Solves for https://github.com/pytorch/pytorch/issues/51080


janeyx99 (approved):

kcarnold:
I'm unsure that this is the correct resolution: if I understand correctly, `11.0.221` is *newer* than `11.0.3`, and `.221` (the newer version) is the problem.

I think it should be pinned at exactly `11.0.3` until some other version is shown to work.

PR #627, opened 2021-01-27
[wip] testing 11.2 binaries

janeyx99:
This will be a draft with all the changes I need to test 11.2 binaries now. Corresponding PR here: https://github.com/pytorch/pytorch/pull/51611

PR #641, opened 2021-02-03
Enable binaries to be built with split torch_cuda 

janeyx99:
Changes check_binary.sh to find the split torch_cuda libraries while refactoring a duplicated chunk into a bash function setup_link_flags.
Edits windows/internal/smoke_test.bat to link against the right libraries when BUILD_SPLIT_CUDA=ON. (This build option was introduced by https://github.com/pytorch/pytorch/pull/49050.)

Disclaimer: does NOT change the cubinsizes.py script to account for the fact that torch_cuda may be split (and thus will cause the libtorch_cuda.so to not be found).

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -230,13 +230,7 @@ else
   done
 fi
 
-TEST_CODE_DIR="$(dirname ${BASH_SOURCE[0]})/test_example_code"
-build_and_run_example_cpp () {
-  if [[ "$DESIRED_DEVTOOLSET" == *"cxx11-abi"* ]]; then
-    GLIBCXX_USE_CXX11_ABI=1
-  else
-    GLIBCXX_USE_CXX11_ABI=0
-  fi
+setup_link_flags () {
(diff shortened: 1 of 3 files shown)


seemethere (approved):

seemethere:
Is there going to be a PR within `pytorch/pytorch` to enable this?

janeyx99:
I tested this code on the CUDA 11.1 binaries in #627 and https://github.com/pytorch/pytorch/pull/47938, but am currently trying to compile CUDA 11.2 binaries without the split (to see if they'd run into relocation marker issues). If they do, I plan to enable the split to alleviate those.

I'm not super confident about this code in particular because of the function refactoring I did in check_binary.sh. I don't think it's wrong but the nightly tests that use check_binary.sh have not passed (example [log](https://app.circleci.com/pipelines/github/pytorch/pytorch/265815/workflows/93e6b2b7-ba78-48a5-bf66-7494487e22d6/jobs/10497347)) due to lack of pre-cxx symbols? I will hold off on merging this until I'm more confident this works.

PR #644, opened 2021-02-03
libtorch: Install cmake=3.14 from conda

seemethere:
cmake from the ubuntu package repositories was too old to actually build
libtorch


janeyx99 (approved):
:D

PR #645, opened 2021-02-04
updating how we get cuda and cudnn version in switch_cuda_version.sh

janeyx99:
For CUDA 11.2, the current cudnn version logic on windows only returns 8, instead of 8.1, which is misleading. Instead, we now use the version defined in cudnn_version.h in the CUDA install, which should be more accurate.

The current cuda version logic also incorrectly returns 110 because cudart.dll in 11.2 (and 11.1) is labeled with 110. I updated this to use nvcc instead, which should be more reliable.

This has been tested in https://github.com/pytorch/pytorch/pull/51611 with CUDA 11.2

### DIFF
--- a/conda/switch_cuda_version.sh
+++ b/conda/switch_cuda_version.sh
@@ -17,13 +17,22 @@ mkdir -p /usr/local
 rm -fr /usr/local/cuda
 ln -s "$CUDA_DIR" /usr/local/cuda
 
+# Using nvcc instead of deducing from cudart version since it's unreliable (was 110 for cuda11.1 and 11.2)
+CUDA_VERSION_DOT=$(nvcc --version | sed -n 4p | cut -f5 -d" " | cut -f1 -d",")
+export CUDA_VERSION=${CUDA_VERSION_DOT/./}
 if [[ "$OSTYPE" == "msys" ]]; then
-    export CUDA_VERSION=`ls /usr/local/cuda/bin/cudart64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
-    export CUDNN_VERSION=`ls /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
+    # we want CUDNN_VERSION=8.1 for CUDA 11.2, not just 8
+    if [[ "$CUDA_VERSION" == '112' ]]; then
+        CUDNN_MAJOR=$(find /usr/local/cuda/ -name cudnn_version.h -exec grep 'define CUDNN_MAJOR' {} + | cut -d' ' -f3)
+        CUDNN_MINOR=$(find /usr/local/cuda/ -name cudnn_version.h -exec grep 'define CUDNN_MINOR' {} + | cut -d' ' -f3)
+        CUDNN_VERSION=$CUDNN_MAJOR.$CUDNN_MINOR
+    else
+        CUDNN_VERSION=$(find /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2)
+    fi
 else
-    export CUDA_VERSION=$(ls /usr/local/cuda/lib64/libcudart.so.*|sort|tac | head -1 | rev | cut -d"." -f -3 | rev)
-    export CUDNN_VERSION=$(ls /usr/local/cuda/lib64/libcudnn.so.*|sort|tac | head -1 | rev | cut -d"." -f -3 | rev)
+    CUDNN_VERSION=$(find /usr/local/cuda/lib64/libcudnn.so.* | sort | tac | head -1 | rev | cut -d"." -f -3 | rev)
 fi
+export CUDNN_VERSION
 
 ls -alh /usr/local/cuda
 


conda/switch_cuda_version.sh
```diff
@@ -17,11 +17,18 @@ mkdir -p /usr/local
 rm -fr /usr/local/cuda
 ln -s "$CUDA_DIR" /usr/local/cuda
 
+CUDA_VERSION_DOT=$(nvcc --version | sed -n 4p|cut -f5 -d" " | cut -f1 -d",")
+export CUDA_VERSION=${CUDA_VERSION_DOT/./}
 if [[ "$OSTYPE" == "msys" ]]; then
-    export CUDA_VERSION=`ls /usr/local/cuda/bin/cudart64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
-    export CUDNN_VERSION=`ls /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
+    export CUDNN_VERSION=$(ls /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2)
```
samestep, line 23: Addressing some Shellcheck warnings:
- [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) Declare and assign separately to avoid masking return values.
- [SC2012](https://github.com/koalaman/shellcheck/wiki/SC2012) Use `find` instead of `ls` to better handle non-alphanumeric filenames.
```suggestion
    CUDNN_VERSION=$(find /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2)
    export CUDNN_VERSION
```

conda/switch_cuda_version.sh
```diff
@@ -17,11 +17,18 @@ mkdir -p /usr/local
...
+
+    # we want CUDNN_VERSION=8.1 for CUDA 11.2, not just 8
+    if [[ "$CUDA_VERSION" == '112' ]]; then
+        CUDNN_MAJOR=`grep 'define CUDNN_MAJOR' $(find /usr/local/cuda/ -name cudnn_version.h) | cut -d' ' -f3`
+        CUDNN_MINOR=`grep 'define CUDNN_MINOR' $(find /usr/local/cuda/ -name cudnn_version.h) | cut -d' ' -f3`
```
samestep, line 28: Addressing some Shellcheck warnings:
- [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006) Use `$(...)` notation instead of legacy backticked `` `...` ``.
- [SC2046](https://github.com/koalaman/shellcheck/wiki/SC2046) Quote this to prevent word splitting
- [SC2038](https://github.com/koalaman/shellcheck/wiki/SC2038) Use -print0/-0 or find -exec + to allow for non-alphanumeric filenames.
```suggestion
        CUDNN_MAJOR=$(find /usr/local/cuda/ -name cudnn_version.h -exec grep 'define CUDNN_MAJOR' {} + | cut -d' ' -f3)
        CUDNN_MINOR=$(find /usr/local/cuda/ -name cudnn_version.h -exec grep 'define CUDNN_MINOR' {} + | cut -d' ' -f3)
```

conda/switch_cuda_version.sh
```diff
@@ -17,11 +17,18 @@ mkdir -p /usr/local
...
+        export CUDNN_VERSION=$CUDNN_MAJOR.$CUDNN_MINOR
+    fi
 else
-    export CUDA_VERSION=$(ls /usr/local/cuda/lib64/libcudart.so.*|sort|tac | head -1 | rev | cut -d"." -f -3 | rev)
     export CUDNN_VERSION=$(ls /usr/local/cuda/lib64/libcudnn.so.*|sort|tac | head -1 | rev | cut -d"." -f -3 | rev)
```
samestep, line 32: Addressing some Shellcheck warnings:
- [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) Declare and assign separately to avoid masking return values.
- [SC2012](https://github.com/koalaman/shellcheck/wiki/SC2012) Use `find` instead of `ls` to better handle non-alphanumeric filenames.
```suggestion
    CUDNN_VERSION=$(find /usr/local/cuda/lib64/libcudnn.so.* | sort | tac | head -1 | rev | cut -d"." -f -3 | rev)
    export CUDNN_VERSION
```

samestep (approved):
Looks good to me overall; could you add a test plan though? I'm not sure how to verify that the changes I suggested are actually correct.

conda/switch_cuda_version.sh
```diff
@@ -17,11 +17,18 @@ mkdir -p /usr/local
...
 if [[ "$OSTYPE" == "msys" ]]; then
-    export CUDA_VERSION=`ls /usr/local/cuda/bin/cudart64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
-    export CUDNN_VERSION=`ls /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2`
+    export CUDNN_VERSION=$(ls /usr/local/cuda/bin/cudnn64*.dll | head -1 | tr '._' ' ' | cut -d ' ' -f2)
+
+    # we want CUDNN_VERSION=8.1 for CUDA 11.2, not just 8
+    if [[ "$CUDA_VERSION" == '112' ]]; then
+        CUDNN_MAJOR=`grep 'define CUDNN_MAJOR' $(find /usr/local/cuda/ -name cudnn_version.h) | cut -d' ' -f3`
+        CUDNN_MINOR=`grep 'define CUDNN_MINOR' $(find /usr/local/cuda/ -name cudnn_version.h) | cut -d' ' -f3`
+        export CUDNN_VERSION=$CUDNN_MAJOR.$CUDNN_MINOR
```
xuzhao9, line 29: Also separate declare and assign here to avoid masking return values.
Is it better to check neither CUDNN_MAJOR nor CUDNN_MINOR is empty before the shell script proceeds?

xuzhao9 (approved):
LGTM!

conda/switch_cuda_version.sh
```diff
@@ -17,11 +17,18 @@ mkdir -p /usr/local
 rm -fr /usr/local/cuda
 ln -s "$CUDA_DIR" /usr/local/cuda
 
+CUDA_VERSION_DOT=$(nvcc --version | sed -n 4p|cut -f5 -d" " | cut -f1 -d",")
```
walterddr, line 20: maybe copy the comment from PR on why using `nvcc --version` --> since nvcc is part of dev not runtime. good to explain that it requires the whole package. not just cuda runtime.

walterddr: also question - did we try using nvidia-smi?

conda/switch_cuda_version.sh
samestep: > Also separate declare and assign here to avoid masking return values.

Shellcheck doesn't warn about that here, my guess is because there's no way for the simple string concatenation to fail.

conda/switch_cuda_version.sh
janeyx99: no I haven't used nvidia-smi

seemethere (approved):

PR #646, opened 2021-02-04
manywheel: Use libcudart.so.11.0 directly for 11.2

seemethere:
libcudart.so.11.0 is actually symlinked to the correct thing so we
should just use it directly since not using it directly appears to cause
failures since the linker is looking for that specific file.


janeyx99 (approved):
oh i had to do this for 11.1 too so looks good

PR #648, opened 2021-02-05
windows: Remove python check from libtorch smoke_test

seemethere:


janeyx99 (commented):
What about the other jobs that use this smoke test like windows conda? Should we add an if instead

seemethere:
> What about the other jobs that use this smoke test like windows conda? Should we add an if instead

Those actually get covered here: https://github.com/pytorch/builder/pull/648/files#diff-59686e8c8da2a64d7f4ddebde6c4e5a0231215eda025e43e4f3311a4ceff27d3L150-L168

janeyx99 (approved):
stellar

PR #652, opened 2021-02-11
[builder] Keep dot in CUDA_VERSION

malfet:
As `pytorch-nightly/build.sh` depends on dot to determine
TORCH_ARCH_LIST


janeyx99 (approved):
looks fine as long as no downstream scripts fail because of this

PR #656, opened 2021-02-13
Cleanup built content after compression on Windows

peterjc123:
To save up more disk space
@janeyx99 @seemethere @malfet


peterjc123:
@seemethere This is also needed to unlock libtorch debug jobs on Windows.

janeyx99 (approved):

PR #657, opened 2021-02-17
conda: Ensure we do download magma correctly

seemethere:
Wasn't actually downloading magma since we added the dot in the
CUDA_VERSION so let's actually get rid of the dot again just for magma.

### DIFF
--- a/conda/pytorch-nightly/bld.bat
+++ b/conda/pytorch-nightly/bld.bat
@@ -11,6 +11,8 @@ if "%USE_CUDA%" == "0" (
 ) else (
     set build_with_cuda=1
     set desired_cuda=%CUDA_VERSION%
+    :: Set up nodot version for use with magma
+    set desired_cuda_nodot=%CUDA_VERSION:.=%
 )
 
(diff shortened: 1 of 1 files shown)


conda/pytorch-nightly/bld.bat
```diff
@@ -11,6 +11,8 @@ if "%USE_CUDA%" == "0" (
 ) else (
     set build_with_cuda=1
     set desired_cuda=%CUDA_VERSION%
+    :: Set up nodot version for use with magma
+    set desired_cuda_nodot=%CUDA_VERSION:.=%
```
seemethere, line 15: cc @peterjc123 just to check that my batch scripting is correct here

walterddr (approved):

conda/pytorch-nightly/bld.bat
seemethere: stolen from, https://www.tutorialspoint.com/batch_script/batch_script_replace_string.htm

peterjc123: @seemethere This is correct.

PR #662, opened 2021-02-22
update package names for both wheel and conda for cross compiled arm64

janeyx99:
Tested at https://github.com/pytorch/pytorch/pull/52624

Attempts to differentiate package names for regular mac builds and arm64 ones by:
- tagging arm64 for conda package version (I tried to look for an architecture variant, but didn't find one). [docs here](https://conda.io/projects/conda-build/en/latest/concepts/package-naming-conv.html#:~:text=Package%20name,are%20referred%20to%20by%20package_name%20.)
- revising the name for wheels, which was easy enough.

### DIFF
--- a/conda/build_pytorch.sh
+++ b/conda/build_pytorch.sh
@@ -385,10 +385,10 @@ for py_ver in "${DESIRED_PYTHON[@]}"; do
         cp "$built_package" "$PYTORCH_FINAL_PACKAGE_DIR/"
     fi
 
-    conda install -y "$built_package"
-
+    # Install the built package and run tests, unless it's for mac cross compiled arm64
     if [[ -z "$CROSS_COMPILE_ARM64" ]]; then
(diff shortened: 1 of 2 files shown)


conda/build_pytorch.sh
```diff
@@ -65,6 +65,12 @@ if [[ -n "$OVERRIDE_PACKAGE_VERSION" ]]; then
     build_version="$OVERRIDE_PACKAGE_VERSION"
     build_number=0
 fi
+
+# testing cross compilation
+if [[ "$CROSS_COMPILE_ARM64" == 1 ]]; then
```
walterddr, line 70: Simply?
```suggestion
if [[ -n "$CROSS_COMPILE_ARM64" ]]; then
```

walterddr (approved):

seemethere (approved):

PR #664, opened 2021-02-23
run_tests: Fix dependency installation for conda tests

seemethere:
Currently when we run `run_tests.sh` with locally installed conda packages the `conda` dependency resolver will sometimes overwrite the local install of `pytorch` with ones from upstream repositories. This is not what we want so to remove the ambiguity I have made a change to utilize `conda update --all` instead which should install dependencies if dependencies are not installed.

_NOTE_: This doesn't work correctly on `cpuonly` packages since they don't actually have a dependency on `cpuonly`, I've included a workaround for that

Example failure: https://app.circleci.com/pipelines/github/pytorch/pytorch/276838/workflows/af4ebc11-3801-4c88-a93a-4dbd1d2103f4/jobs/11093808

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -61,35 +61,24 @@ fi
 
 # Environment initialization
 if [[ "$package_type" == conda || "$(uname)" == Darwin ]]; then
-    EXTRA_CONDA_FLAGS=""
-    if [[ "$(python --version 2>&1)" == *3.9.* || "$DESIRED_CUDA" == cu111 || "$DESIRED_CUDA" == cu112 ]]; then
-      EXTRA_CONDA_FLAGS="-c=conda-forge"
-    fi
-
-    # Warning: if you've already manually installed the built conda package
-    # your environment is probably inconsistent (because most of the packages
-    # have a feature requirement).  If we post-facto install the feature,
-    # that will make the environment consistent again.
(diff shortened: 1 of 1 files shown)


run_tests.sh
```diff
@@ -61,35 +61,24 @@ fi
...
-        fi
-    fi
-    if [[ "$(python --version 2>&1)" == *3.9.* ]]; then
-        retry conda install -yq ${EXTRA_CONDA_FLAGS} future hypothesis ninja protobuf pytest setuptools six typing_extensions pyyaml
-    elif [[ "$(python --version 2>&1)" == *3.8.* ]]; then
-        retry conda install -yq future hypothesis mkl>=2018 ninja numpy>=1.15 protobuf pytest setuptools six typing_extensions pyyaml
-    elif [[ "$(python --version 2>&1)" == *3.6.* ]]; then
-        retry conda install -yq cffi future hypothesis mkl>=2018 ninja numpy>=1.11 protobuf pytest setuptools six typing_extensions pyyaml requests dataclasses
+    EXTRA_CONDA_FLAGS="-c=conda-forge"
+    # Why is there two different ways to install dependencies after installing an offline package?
```
janeyx99, line 65:
```suggestion
    # Why are there two different ways to install dependencies after installing an offline package?
```
nit

janeyx99 (approved):
wow we love conda so much

PR #668, opened 2021-02-26
manywheel: Drop 7.5 for CUDA 10.2

seemethere:
sm_75 should be forward compatible with sm_70 so dropping it shouldn't
make much of a difference while giving us 173MB to work with

### DIFF
--- a/manywheel/build.sh
+++ b/manywheel/build.sh
@@ -70,7 +70,7 @@ case ${CUDA_VERSION} in
         EXTRA_CAFFE2_CMAKE_FLAGS+=("-DATEN_NO_TEST=ON")
         ;;
     10.*)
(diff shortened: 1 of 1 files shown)


nelson-liu:
@seemethere , if you don't mind me asking, why only remove 7.5 for CUDA 10.2, and not 11.x as well?

seemethere:
> @seemethere , if you don't mind me asking, why only remove 7.5 for CUDA 10.2, and not 11.x as well?

We currently only publish CUDA 10.2 on PyPI where we have a size restriction of 750MB for binaries.

When we start publishing CUDA 11.X on PyPI we’ll probably go through a similar exercise of pruning architectures

PR #671, opened 2021-03-03
Add support for building aarch64-wheels in Docker

malfet:
Refactor existing codebase to have RemoteHost abstraction
After that, most of the build script will work without a change inside or outside of docker
Add "--use-docker" option to build true manylinux2014 compatible wheels inside official pypa docker

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -274,6 +274,8 @@ def start_build(host: RemoteHost, *,
     print('Checking out TorchVision repo')
     if branch.startswith("v1.7.1"):
         host.run_cmd(f"git clone https://github.com/pytorch/vision -b v0.8.2-rc2 {git_clone_flags}")
+    if branch.startswith("v1.8.0"):
+        host.run_cmd(f"git clone https://github.com/pytorch/vision -b v0.9.0-rc3 {git_clone_flags}")
     else:
         host.run_cmd(f"git clone https://github.com/pytorch/vision {git_clone_flags}")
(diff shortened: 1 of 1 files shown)


build_aarch64_wheel.py
```diff
@@ -82,121 +145,151 @@ def wait_for_connection(addr, port, timeout=5, attempt_cnt=5):
...
-    run_ssh(addr, "sudo apt-get install -y ocaml ocamlbuild autoconf automake indent libtool fig2dev texinfo")
-    # TODO: fix a version to build
-    # TODO: consider adding flags --host=arm-linux-gnueabi --enable-single --enable-neon CC=arm-linux-gnueabi-gcc -march=armv7-a -mfloat-abi=softfp
-    run_ssh(addr, f"git clone https://github.com/FFTW/fftw3 {git_clone_flags}")
-    run_ssh(addr, "pushd fftw3; sh bootstrap.sh; make -j8; sudo make install; popd")
+        host.run_cmd("sudo pip3 install Cython")
+        host.run_cmd("sudo pip3 install numpy")
+
+    build_OpenBLAS(host, git_clone_flags)
+    # build_FFTW(host, git_clone_flags)
```
walterddr, line 237: are we disabling fftw?

walterddr (approved):
lgtm

build_aarch64_wheel.py
malfet: Because it is not used from the codebase at the moment, is it?

PR #682, opened 2021-03-23
Make aarch64 wheels RHEL8 compatible

malfet:
RedHat-8 on aarch64 expects all loadable segments of shared libraries to be 64KB aligned.
Modify build_aarch64_wheel.py to:
 - Pass `-Wl,-z,max-page-size=0x10000` to the linker flags
 - Ensure that page-alignment is not change when patchelf is running
 - Add option to allocate rhel8 machine
 - Assign default tags to allocated machines
 
Fixes https://github.com/pytorch/pytorch/issues/53597

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -9,8 +9,12 @@ from typing import List, Optional, Tuple, Union
 
 
 # AMI images for us-east-1, change the following based on your ~/.aws/config
-ubuntu18_04_ami = "ami-0f2b111fdc1647918"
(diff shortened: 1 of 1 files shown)


walterddr (approved):
looks good to me. 
Although I am not sure if we can just make the image align to 64kB unconditionally. seems like it is backward compatible with 4kB alignment setting right?

malfet:
I've tried it (that's the reason why I've added redhat8 allocation option) and it works fine. And yes 64Kb aligned segments are compatible with OS flavors using 4Kb pages

PR #684, opened 2021-03-24
Update MSVC toolchain selection for Windows builds

peterjc123:


peterjc123:
cc @seemethere @malfet

malfet (approved):

windows/internal/vc_install_helper.bat
```diff
@@ -5,8 +5,28 @@ if "%VC_YEAR%" == "2017" if "%CUDA_VERSION%" == "92" (
...
 )
+if "%VC_YEAR%" == "2019" set VC_VERSION=14.28.29333
 if not "%VC_VERSION%" == "" (
     set VSDEVCMD_ARGS=-vcvars_ver=%VC_VERSION%
 )
 if "%VC_YEAR%" == "2017" powershell windows/internal/vs2017_install.ps1 %VC_VERSION%
+if "%VC_YEAR%" == "2019" powershell windows/internal/vs2019_install.ps1 %VC_VERSION%
+
+set VC_VERSION_LOWER=16
+set VC_VERSION_UPPER=17
```
mszhanyi, line 16: If we don't use VS2017, it'd better to remove them all

peterjc123: Let's do that in a follow-up PR since this is not the only one spot that involves VS 2017.

seemethere (approved):

PR #686, opened 2021-03-25
Use packaging.Version to compare version numbers instead of direct string comparison

xuzhao9:
This fixes the recent torchtext nightlies with version 0.10.x to be incorrectly hidden.

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -7,6 +7,7 @@ from os import path
 from collections import defaultdict
 from typing import List, Type, Dict, Set, TypeVar, Generator, Optional
 from re import sub, match
+from packaging.version import parse
 
 import botocore
 import boto3
@@ -61,6 +62,7 @@ class S3Index:
         # sorting, sorts in reverse to put the most recent versions first
         all_sorted_packages = sorted(
             {self.normalize_package_version(obj) for obj in self.objects},
+            key=lambda v: parse(v.split('-')[1]),
             reverse=True,
         )
         packages: Dict[str, int] = defaultdict(int)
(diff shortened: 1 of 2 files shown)


s3_management/manage.py
```diff
@@ -61,6 +62,7 @@ def nightly_packages_to_show(self: S3IndexType) -> Set[str]:
         # sorting, sorts in reverse to put the most recent versions first
         all_sorted_packages = sorted(
             {self.normalize_package_version(obj) for obj in self.objects},
+            key=lambda v: parse(v.split('-')[1]),
```
janeyx99, line 65: Wouldn't parse(v) work as well/with more precision?

s3_management/manage.py
```diff
@@ -7,6 +7,7 @@
 from collections import defaultdict
 from typing import List, Type, Dict, Set, TypeVar, Generator, Optional
 from re import sub, match
+from packaging.version import parse
```
seemethere, line 10: Is this from a third party library?

s3_management/manage.py
malfet, line 65: Can we do something like that to make it more resilient to errors. 
```suggestion
            key=lambda v: parse(v.split('-', 1)[-1]),
```

janeyx99: Oh sorry, never mind. I didn't realize v was not just the version, but the entire package name!

malfet (approved):

s3_management/manage.py
malfet: It's https://github.com/pypa/packaging (which is dependency for wheel)

s3_management/manage.py
xuzhao9: Yeah the naming is a bit confusing here. I have renamed the lambda variable.

xuzhao9: They should have the same effect because we only use the first two parts of split here: https://github.com/pytorch/builder/blob/master/s3_management/manage.py#L85

s3_management/manage.py
xuzhao9: Yes, I have added the dependency to requirements.txt. Thanks for pointing it out.

PR #692, opened 2021-04-16
Add llvm-9.0.1 to the docker images

bertmaher:
We're moving towards enabling the CPU compiler backend, which depends
on LLVM.  Locally I measured that this adds 11 MB to the wheel (27 for
the uncompressed libraries).

I also checked that it coexists nicely with numba/llvmlite (e.g., no
symbol conflicts when both are loaded).


bertmaher:
I apparently can't add reviewers, but cc @malfet @seemethere

malfet (approved):

seemethere (approved):

PR #701, opened 2021-04-20
Build nightly pytorch with llvm included

bertmaher:
Enable LLVM in the nightly build.  This shouldn't enabled anything by default, but should make the LLVM-based CPU fuser available so that we can kick the proverbial tires.

### DIFF
--- a/conda/pytorch-nightly/build.sh
+++ b/conda/pytorch-nightly/build.sh
@@ -6,6 +6,8 @@ export CMAKE_PREFIX_PATH=$PREFIX
 export TH_BINARY_BUILD=1 # links CPU BLAS libraries thrice in a row (was needed for some MKL static linkage)
 export PYTORCH_BUILD_VERSION=$PKG_VERSION
 export PYTORCH_BUILD_NUMBER=$PKG_BUILDNUM
+export USE_LLVM="/opt/llvm_no_cxx11_abi"
+export LLVM_DIR="$USE_LLVM/lib/cmake/llvm"
 
 # set OPENSSL_ROOT_DIR=/opt/openssl if it exists
 if [[ -e /opt/openssl ]]; then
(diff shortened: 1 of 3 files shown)


bertmaher:
I haven't figured out how to test this locally so I'm sort of guessing that these are the right places to add the build variables.

wheel/build_wheel.sh
```diff
@@ -134,6 +134,8 @@ export TH_BINARY_BUILD=1
 export INSTALL_TEST=0 # dont install test binaries into site-packages
 export MACOSX_DEPLOYMENT_TARGET=10.10
 export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
+export USE_LLVM="/opt/llvm"
```
seemethere, line 137: These scripts are mainly used for mac wheels which probably won't use this feature?

bertmaher: I think using this for mac is reasonable, llvm is pretty well supported there, the CPU backend is probably useful there, and the nnc team has the ability to test it locally.  So I was going to try to land it there unless it really complicates things.

bertmaher: Per @malfet I can leave this off for Mac so we get coverage of a non-llvm enabled build.  I think the Linux builds are the ones I most want to have this anyways.

bertmaher:
@seemethere @malfet I believe this is ready to go!  I kicked off the binary builds using the ci/binaries label on https://github.com/pytorch/pytorch/pull/56637; I'm not 100% sure what's expected to succeed there (a number of the builds failed with "Build was canceled"), but the builds I expect to work do, and the wheel size looks acceptable (e.g. binary_linux_manywheel_3_9m_cu102_devtoolset7 is 833 MB: https://app.circleci.com/pipelines/github/pytorch/pytorch/358987/workflows/53f0ee72-d84e-49bd-ad18-1c8ae804e2f6/jobs/15142617).

Additionally I verified by hand that the feature actually works by downloading the wheel and running a few tests.

Shall we ship this and see how it goes?

seemethere (approved):

malfet:
@bertmaher rolling it back, as it caused crashes when installed in conjunction with librosa. (see  https://github.com/pytorch/pytorch/issues/62754) I guess PR needs to be reworked to build the library with `-fvisibility=hidden`.

PR #705, opened 2021-04-22
Disable ffmpeg_mac and ffmpeg_win jobs

driazati:
These are broken and aren't being monitored anyways, so delete them

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,29 +173,16 @@ workflows:
           cu_version: cpu
           is-python3: true
           python_version: '3.7'
-      - ffmpeg_mac:
-          cu_version: cpu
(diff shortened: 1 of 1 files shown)


seemethere (approved):

seemethere:
I think it might also be good to just remove the ffmpeg builds altogether since they haven't been actually been maintained for a while. cc @andfoy

andfoy:
We're still blocked by https://github.com/pytorch/vision/pull/2818, however FFmpeg 4.4 was released a few days ago, I don't know if we should build that version. @fmassa @bjuncek

seemethere:
In any case I'm going to go ahead and merge this to disable the mac builds for now

PR #727, opened 2021-04-28
Use torch_cpu specific debug flag

driazati:
See https://github.com/pytorch/pytorch/pull/57190 for details

### DIFF
--- a/manywheel/build_libtorch.sh
+++ b/manywheel/build_libtorch.sh
@@ -129,7 +129,7 @@ fi
         EXTRA_CAFFE2_CMAKE_FLAGS="${EXTRA_CAFFE2_CMAKE_FLAGS[@]} $STATIC_CMAKE_FLAG" \
         # TODO: Remove this flag once https://github.com/pytorch/pytorch/issues/55952 is closed
         CFLAGS='-Wno-deprecated-declarations' \
-        REL_WITH_DEB_INFO=1 \
+        BUILD_LIBTORCH_CPU_WITH_DEBUG=1 \
         python setup.py install
 
     mkdir -p libtorch/{lib,bin,include,share}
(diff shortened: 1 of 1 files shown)


manywheel/build_libtorch.sh
```diff
@@ -155,16 +155,14 @@ fi
...
 
+    # Remove debug info from release lib
+    strip --strip-debug libtorch/lib/libtorch_cpu.so
+
     # Add a debug link to the release lib to the debug lib (debuggers will then
     # search for symbols in a file called libtorch_cpu.so.dbg in some
     # predetermined locations) and embed a CRC32 of the debug library into the .so
     cd libtorch/lib
 
-    # Clean out debug symbols from all the libs
```
driazati, line 163: These are removed since we only need to strip the libraries built with debug info (before it was everything, now it's just `libtorch_cpu.so`, which gets stripped above).

[Binary sizes look OK](https://app.circleci.com/pipelines/github/pytorch/pytorch/312690/workflows/7f18f397-2150-4447-9b3d-d9e4cbd310e0/jobs/12976773):

```
total 1.1G
drwxr-xr-x 2 root root  183 Apr 29 01:21 .
drwxr-xr-x 1 root root  215 Apr 29 01:19 ..
-rw-r--r-- 1 root root 662M Apr 29 01:21 debug-libtorch-shared-with-deps-1.9.0.dev20210429+cpu-e4cc6f8d.zip
-rw-r--r-- 1 root root 183M Apr 29 01:21 libtorch-shared-with-deps-1.9.0.dev20210429+cpu.zip
-rw-r--r-- 1 root root 183M Apr 29 01:21 libtorch-shared-with-deps-latest.zip
```

seemethere (approved):

PR #730, opened 2021-04-28
Remove conda-forge from our testing scripts

seemethere:
We shouldn't be relying on conda-forge for our testing purposes since it
contains packages that can conflict with the current way we build
pytorch.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -61,7 +61,6 @@ fi
 
 # Environment initialization
 if [[ "$package_type" == conda || "$(uname)" == Darwin ]]; then
-    EXTRA_CONDA_FLAGS="-c=conda-forge"
     # Why are there two different ways to install dependencies after installing an offline package?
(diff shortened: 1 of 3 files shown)


run_tests.sh
```diff
@@ -61,7 +61,6 @@ fi
 
 # Environment initialization
 if [[ "$package_type" == conda || "$(uname)" == Darwin ]]; then
-    EXTRA_CONDA_FLAGS="-c=conda-forge"
```
malfet, line 64: I think we still need to keep it for Darwin and Python-3.9, aren't we?

malfet (approved):

run_tests.sh
seemethere: Just tried the following and it appears to work fine,

```
❯ conda create -y -p $(pwd) -c pytorch pytorch python=3.9
```

PR #749, opened 2021-04-29
.github: Add workflows to build conda/libtorch images

seemethere:

### DIFF
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -1,0 +1,26 @@
+name: Build conda docker images
+
+on:
+  pull_request:
+    paths:
+      - conda/Dockerfile
+      - 'common/*'
+
+env:
+  DOCKER_REGISTRY: "docker.io"
+  DOCKER_BUILDKIT: 1
+
+jobs:
+  build-docker:
+    runs-on: linux.2xlarge
+    strategy:
+      matrix:
+        cuda_version: ["10.2", "11.1", "11.3", "cpu"]
+    env:
+      CUDA_VERSION: ${{ matrix.cuda_version }}
+    steps:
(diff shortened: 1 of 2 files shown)


common/install_magma.sh
```diff
@@ -2,6 +2,7 @@
 
 set -eou pipefail
 
+# Check https://anaconda.org/pytorch/magma-cuda${cuda_version_nodot} to see if there's a magma version available
```
walterddr, line 5: is this an instruction to developer when changing the MAGMA_VERSION below?

walterddr (approved):

.github/workflows/build-conda-images.yml
```diff
@@ -0,0 +1,26 @@
...
+      matrix:
+        cuda_version: ["10.2", "11.1", "11.3", "cpu"]
+    env:
+      CUDA_VERSION: ${{ matrix.cuda_version }}
+    steps:
+      - name: Checkout PyTorch
+        uses: actions/checkout@v2
+      - name: Build Docker Image
+        run: |
+          conda/build_docker.sh
```
janeyx99, line 26: So these images get built when there are changes within the builder repo, right? This wouldn't be replacing the circle docker jobs just yet?

janeyx99 (approved):

malfet (approved):

common/install_magma.sh
seemethere: This is just a comment to trigger the workflows, I'll probably drop this commit once I verify that everything's run

.github/workflows/build-conda-images.yml
seemethere: Yes these are only for the docker images we currently use to build the binary builds

PR #755, opened 2021-05-04
Add thrust_11.3 explicitly and remove bad quotes

janeyx99:
thrust_11.3 was moved into its separate package so we need to explicitly put it as an argument.

The quotes break installation, so removing them as they're not actually necessary

### DIFF
--- a/windows/internal/smoke_test.bat
+++ b/windows/internal/smoke_test.bat
@@ -2,10 +2,7 @@ set SRC_DIR=%~dp0
 
 pushd %SRC_DIR%\..
 
-if "%CUDA_VERSION%" == "102" call internal\driver_update.bat
-if "%CUDA_VERSION%" == "110" call internal\driver_update.bat
-if "%CUDA_VERSION%" == "111" call internal\driver_update.bat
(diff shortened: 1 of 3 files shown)


seemethere (approved):

windows/internal/smoke_test.bat
```diff
@@ -2,10 +2,7 @@ set SRC_DIR=%~dp0
 
 pushd %SRC_DIR%\..
 
-if "%CUDA_VERSION%" == "102" call internal\driver_update.bat
-if "%CUDA_VERSION%" == "110" call internal\driver_update.bat
-if "%CUDA_VERSION%" == "111" call internal\driver_update.bat
-if "%CUDA_VERSION%" == "112" call internal\driver_update.bat
+if DEFINED CUDA_VERSION call internal\driver_update.bat
```
seemethere, line 5: Couldn't CUDA_VERSION also be set to cpu?

malfet (approved):

windows/internal/smoke_test.bat
janeyx99: Good call, I'll change this to check for inequality with cpu

PR #760, opened 2021-05-05
add call to build_amd.py to build_libtorch.sh

jeffdaily:
Similar to use of build_amd.py in manywheel/build_common.sh.

### DIFF
--- a/manywheel/build_libtorch.sh
+++ b/manywheel/build_libtorch.sh
@@ -115,6 +115,11 @@ else
     export _GLIBCXX_USE_CXX11_ABI=0
 fi
 
+if [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
+    echo "Calling build_amd.py at $(date)"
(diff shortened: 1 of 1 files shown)


manywheel/build_libtorch.sh
```diff
@@ -115,6 +115,11 @@ else
     export _GLIBCXX_USE_CXX11_ABI=0
 fi
 
+if [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
+    echo "Calling build_amd.py at $(date)"
+    python tools/amd_build/build_amd.py
```
malfet, line 120: Do we need to exit after calling build_amd.py? Or continue execution?

malfet (approved):

manywheel/build_libtorch.sh
jeffdaily: Continue execution, yes.  This is the step that hipifies the pytorch sources.

PR #761, opened 2021-05-05
[Conda] convert osx-arm64 packages

malfet:
Conda does not correctly understand the way our cross-compilation is set up
Fix that by explicitly calling `conda convert -p osx-arm64`

Also delete numpy install dependency (as it is dynamic) and make `mkl` dependency x86_64 only


malfet:
Testing it here: https://app.circleci.com/pipelines/github/pytorch/pytorch/316414/workflows/631c2814-9f2e-4ae8-9ddb-825dd267004f/jobs/13191080

seemethere (approved):

PR #763, opened 2021-05-10
Test using ninja for manylinux

janeyx99:


janeyx99:
tests done--ninja is preferable

PR #766, opened 2021-05-14
Add duplicates_analyze script

malfet:
Prints duplicated symbols between various libraries

For example:
```
build/lib/libtorch_python.dylib symbols size 15.81 Mb libshm.dylib overlap is 2 Kb libtorch_cpu.dylib overlap is 1.12 Mb libc10.dylib overlap is 32 Kb
build/lib/libshm.dylib symbols size 31 Kb libtorch.dylib overlap is 0 bytes libtorch_cpu.dylib overlap is 2 Kb libc10.dylib overlap is 2 Kb
build/lib/libtorch.dylib symbols size 0 bytes
build/lib/libtorch_cpu.dylib symbols size 103.98 Mb libc10.dylib overlap is 134 Kb
```

### DIFF
--- a/analytics/duplicates_analyze.py
+++ b/analytics/duplicates_analyze.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+from typing import Dict, List
+from subprocess import check_output
+import os
+import sys
+
+def get_defined_symbols(fname: str) -> Dict[str, int]:
+    if sys.platform == 'darwin':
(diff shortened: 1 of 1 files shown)


seemethere (approved):

driazati (approved):
lgtm, for later it'd be cool to see the actual duplicated symbols (probably [demangled](https://pypi.org/project/cxxfilt/) too) and also have JSON output so we can dump this into S3 at regular intervals

PR #767, opened 2021-05-14
Install ninja in manywheel images

janeyx99:
This way we can use ninja in our binary builds for manywheel, based on instructions from: https://centos.pkgs.org/7/okey-x86_64/ninja-build-1.8.2-1.el7.x86_64.rpm.html

### DIFF
--- a/.github/workflows/build-manywheel-images.yml
+++ b/.github/workflows/build-manywheel-images.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: linux.2xlarge
     strategy:
       matrix:
-        cuda_version: ["10.2", "11.1"]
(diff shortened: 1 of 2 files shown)


malfet (approved):

qhaas:
`ninja-build` is now in the [epel7 repository](https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/n/) and now uses the [expected invocation command](https://bugzilla.redhat.com/show_bug.cgi?id=1608565).  Given epel is a arguably a bit more mainstream than OKay, would it not make sense just to use the one from epel7 without leveraging an additional third-party repo?

PR #768, opened 2021-05-17
enable credentials for pushing manylinux images

janeyx99:
Adds a mechanism so that the build jobs will push the images when the PR is merged to master.

Uses secrets DOCKER_ID and DOCKER_TOKEN for Docker Hub.

### DIFF
--- a/.github/workflows/build-manywheel-images.yml
+++ b/.github/workflows/build-manywheel-images.yml
@@ -1,6 +1,9 @@
 name: Build manywheel docker images
 
 on:
+  push:
+    branches:
+      master
   pull_request:
     paths:
       - manywheel/Dockerfile
@@ -9,6 +12,9 @@ on:
 env:
   DOCKER_REGISTRY: "docker.io"
   DOCKER_BUILDKIT: 1
+  DOCKER_ID: ${{ secrets.DOCKER_ID }}
+  DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
+  WITH_PUSH: ${{ github.event_name == 'push' }}
 
 jobs:
   build-docker-cuda:
--- a/.github/workflows/build-manywheel-images.yml
+++ b/.github/workflows/build-manywheel-images.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       master
+    paths:
+      - manywheel/Dockerfile
+      - 'common/*'
   pull_request:
     paths:
       - manywheel/Dockerfile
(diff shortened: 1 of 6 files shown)


.github/workflows/build-manywheel-images.yml
```diff
@@ -1,6 +1,9 @@
 name: Build manywheel docker images
 
 on:
+  push:
```
samestep, line 4: do you need to specify `paths` here too?
```suggestion
  push:
    paths:
      - manywheel/Dockerfile
```

manywheel/build_docker.sh
```diff
@@ -57,9 +57,10 @@ DOCKER_IMAGE=${DOCKER_REGISTRY}/pytorch/manylinux-builder:${DOCKER_TAG}
     docker tag ${DOCKER_IMAGE} ${LEGACY_DOCKER_IMAGE}
 )
 
-if [[ -n ${WITH_PUSH} ]]; then
+if [[ "${WITH_PUSH}" == true ]]; then
     (
         set -x
+        echo "${DOCKER_TOKEN}" | docker login -u "${DOCKER_ID}" --password--stdin
```
seemethere, line 63: Let's try and have the login just be a separate step in the workflow, we don't want to have users to have to provide these environment variables in order to actually run this script

manywheel/build_docker.sh
```diff
@@ -57,9 +57,10 @@ DOCKER_IMAGE=${DOCKER_REGISTRY}/pytorch/manylinux-builder:${DOCKER_TAG}
     docker tag ${DOCKER_IMAGE} ${LEGACY_DOCKER_IMAGE}
 )
 
-if [[ -n ${WITH_PUSH} ]]; then
+if [[ "${WITH_PUSH}" == true ]]; then
```
seemethere, line 60: What's the reasoning behind this change? Wouldn't it also just work without needing to change the actual equals value here?

.github/workflows/build-manywheel-images.yml
janeyx99: oh good call

manywheel/build_docker.sh
janeyx99: It's cuz I set it to ${{ github.event_name == 'push' }}, which I think would be 'false' if we're not pushing, which would pass the `-n`, no?

seemethere: Oh that's true, we should probably document this in the README.md

Perhaps we can also make this change for the rest of the docker build scripts as well?

seemethere (approved):

Issue #772, opened 2021-05-20
Create backup tags for binary build images


seemethere:
Now that we are building the binary build images within CI and pushing them within CI we should create an automatic backup tag of the image that we're building that's based on the commit sha and branch reference.

Example:

1. We build and push `pytorch/conda-builder:cpu`
2. We tag `pytorch/conda-builder:cpu` as both `pytorch/conda-builder:cpu-${GIT_BRANCH_NAME}` and `pytorch/conda-builder:cpu-${GIT_COMMIT_SHA}`
3. Push all tagged images to docker hub

Tags should look somewhat similar to:
```
pytorch/conda-builder:cpu
pytorch/conda-builder:cpu-master
pytorch/conda-builder:cpu-jkfdl1234j113uy341hj
```

This should be applied to all three of:
- [ ] pytorch/conda-builder
- [ ] pytorch/manylinux-builder
- [ ] pytorch/libtorch-cxx11-builder

PR #773, opened 2021-05-20
Add tags to manywheel; fix typo

janeyx99:
A step towards fixing https://github.com/pytorch/builder/issues/772.

Also, my latest PR did not push successfully (https://github.com/pytorch/builder/runs/2633758499?check_suite_focus=true). This would fix that (I tested the login command locally this time with Docker Desktop)


manywheel/build_docker.sh
```diff
@@ -52,16 +52,30 @@ DOCKER_IMAGE=${DOCKER_REGISTRY}/pytorch/manylinux-builder:${DOCKER_TAG}
         "${TOPDIR}"
 )
 
+GITHUB_REF=${GITHUB_REF:-}
+GIT_BRANCH_NAME=${GITHUB_REF##*/}
+GIT_COMMIT_SHA=${GITHUB_SHA:-}
+DOCKER_IMAGE_BRANCH_TAG=${DOCKER_IMAGE}-${GIT_BRANCH_NAME}
+DOCKER_IMAGE_SHA_TAG=${DOCKER_IMAGE}-${GIT_COMMIT_SHA}
+
```
seemethere, line 60: Can we set the default values to something that we can derive locally here?

i.e. `GITHUB_REF=$(git symbolic-ref -q --short HEAD || git describe --tags --exact-match)`

```suggestion
GITHUB_REF=${GITHUB_REF:$(git symbolic-ref -q --short HEAD || git describe --tags --exact-match)}
GIT_BRANCH_NAME=${GITHUB_REF##*/}
GIT_COMMIT_SHA=${GITHUB_SHA:-$(git rev-parse HEAD)}
DOCKER_IMAGE_BRANCH_TAG=${DOCKER_IMAGE}-${GIT_BRANCH_NAME}
DOCKER_IMAGE_SHA_TAG=${DOCKER_IMAGE}-${GIT_COMMIT_SHA}

```

Just to make it so we can run this locally as well and don't have to rely on the env variables being present

seemethere (changes requested):

manywheel/build_docker.sh
janeyx99: done!

seemethere (approved):

PR #777, opened 2021-05-25
Add some default values to environment variables

malfet:
Make it easier to debug local builds, one just need to specify `PYTORCH_GITHUB_ROOT_DIR`


janeyx99 (approved):
yayyy more defaults!!

Issue #670, opened 2021-03-01
Target manylinux_2_24 for wheels


mattip:
The new manylinux_2_24 standard, based on debian9 and glibc2.24, could enable #520 and solve pytorch/pytorch#51039. It would require a new [Dockerfile](https://github.com/pytorch/builder/blob/master/manywheel/Dockerfile), changing all the `yum` installation calls to `apt`, and changing the PATCHELF parts of the build.

mattip:
Copying selected comments from pypa/manylinux#1012, which discusses the fact that `manylinux_2_24` is based on Ubuntu, which does not recompile newer gcc versions for the `libstdc++` needed for `manylinux_2_24` compliance:

------------------------------
> The GCC version in `manylinux_2_24` is a dismal 6.3, much worse than both the year based manylinuxes and not even new enough for C++17 language support (mostly added in GCC 7)! This is a huge step backward for an image with a higher GLIBC version.

> Once you install gcc-9, it will update at least libstdc++ which is a no go. RHEL dev toolset took care of that and also other libraries like libgcc_s so that binaries produced with devtoolset would still be compatible with the base image with no action.

------------------------

Bottom line: in order to use the `manylinux_2_24` standard for wheels (which would enable using `CXX11_ABI`), someone needs to put in the effort to build a gcc-9 or gcc-10 for `manylinux_2_24`.

rgommers:
To add to this:
- currently we still have `manylinux1` wheels - that means `glibc` 2.5 (2006)
- upgrading to `manylinux2014` is much easier - that would be `glibc` 2.17 (2012)
- `manylinux_2_24` gives `glibc` 2.24 (2016)

@mattip you said 2.5 -> 2.17 was much less interesting than 2.17 -> 2.24. Can you explain why that is? Newer is better, but getting rid of `manylinux1` would also be interesting I'd think?

mattip:
CentOS7, which is the base distro for `manylinux2014`, is the reason pytorch cannot use CXX11_ABI. Moving to `manylinux2014` is easy and makes sense since the older distros are EOL, but I am not sure what the tangible benefits are for pytorch.

For a nice summary of glibc versions vs. distros, along with EOL dates, there is https://github.com/mayeut/pep600_compliance

snnn:
[...]

PR #785, opened 2021-06-14
Add manywheel2014-builder images

mattip:
This PR
- adds a Dockerfile_2014 that uses the pre-build manylinux2014 image as a base, meaning no need to build python, auditwheel, and patchelf
- tries to remain true to the previous docker image so that building pytorch should JustWork
- convert the build script into a loop for both docker image bases, which should create two sets of docker images

If this looks OK and actually uploads some images, I think the next iterative steps would be a PR to use the images (tagged with `manylinux2014-builder:XXX`) in pytorch/pytorch, and then to come back and fix the image failures.

### DIFF
--- a/manywheel/Dockerfile_2014
+++ b/manywheel/Dockerfile_2014
@@ -1,0 +1,143 @@
+# syntax = docker/dockerfile:experimental
+ARG ROCM_VERSION=3.7
+ARG BASE_CUDA_VERSION=10.1
+ARG GPU_IMAGE=nvidia/cuda:${BASE_CUDA_VERSION}-devel-centos7
+FROM quay.io/pypa/manylinux2014_x86_64 as base
+
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+
+RUN yum install -y wget curl perl util-linux xz bzip2 git patch which perl zlib-devel
+RUN yum install -y yum-utils centos-release-scl
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms
+RUN yum install -y devtoolset-7-gcc devtoolset-7-gcc-c++ devtoolset-7-gcc-gfortran devtoolset-7-binutils
+ENV PATH=/opt/rh/devtoolset-7/root/usr/bin:$PATH
+ENV LD_LIBRARY_PATH=/opt/rh/devtoolset-7/root/usr/lib64:/opt/rh/devtoolset-7/root/usr/lib:$LD_LIBRARY_PATH
+
+# cmake
+RUN yum install -y cmake3 && \
+    ln -s /usr/bin/cmake3 /usr/bin/cmake
+FROM base as openssl
+# Install openssl (this must precede `build python` step)
+# (In order to have a proper SSL module, Python is compiled
+# against a recent openssl [see env vars above], which is linked
+# statically. We delete openssl afterwards.)
+ADD ./common/install_openssl.sh install_openssl.sh
+RUN bash ./install_openssl.sh && rm install_openssl.sh
+
+
+
+# remove unncessary python versions
+RUN rm -rf /opt/python/cp26-cp26m /opt/_internal/cpython-2.6.9-ucs2
+RUN rm -rf /opt/python/cp26-cp26mu /opt/_internal/cpython-2.6.9-ucs4
+RUN rm -rf /opt/python/cp33-cp33m /opt/_internal/cpython-3.3.6
+RUN rm -rf /opt/python/cp34-cp34m /opt/_internal/cpython-3.4.6
+
+FROM base as cuda
+ARG BASE_CUDA_VERSION=10.1
+# Install CUDA
+ADD ./common/install_cuda.sh install_cuda.sh
+RUN bash ./install_cuda.sh ${BASE_CUDA_VERSION} && rm install_cuda.sh
+
+FROM base as intel
+# MKL
+ADD ./common/install_mkl.sh install_mkl.sh
+RUN bash ./install_mkl.sh && rm install_mkl.sh
+
+FROM base as magma
+ARG BASE_CUDA_VERSION=10.1
+# Install magma
+ADD ./common/install_magma.sh install_magma.sh
+RUN bash ./install_magma.sh ${BASE_CUDA_VERSION} && rm install_magma.sh
+
+FROM base as jni
+# Install java jni header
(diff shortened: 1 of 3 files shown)


mattip:
A first step toward #670 which

manywheel/build_docker.sh
```diff
@@ -40,42 +40,59 @@ case ${GPU_ARCH_TYPE} in
...
-    DOCKER_BUILDKIT=1 docker build \
-        -t "${DOCKER_IMAGE}" \
-        ${DOCKER_GPU_BUILD_ARG} \
-        --build-arg "GPU_IMAGE=${GPU_IMAGE}" \
-        --target "${TARGET}" \
-        -f "${TOPDIR}/manywheel/Dockerfile" \
-        "${TOPDIR}"
-)
+IMAGES=''
+for DOCKER_NAME in manylinux manylinux2014; do
```
seemethere, line 44: I think if we're going to do this it should just be an env variable that you pass through instead of iterating a `for` loop here, something like:

```
MANYLINUX_VERSION=2014 manywheel/build_docker.sh
```

Just so that we don't have to build all of the variants every single time.

This logic to build all of the variants can be added to `manywheel/build_all_docker.sh`

mattip: adopted

mattip:
A maintainer needs to approve the CI run

mattip:
ping @seemethere

manywheel/Dockerfile_2014
```diff
@@ -0,0 +1,143 @@
...
+
+
+# remove unncessary python versions
+RUN rm -rf /opt/python/cp26-cp26m /opt/_internal/cpython-2.6.9-ucs2
+RUN rm -rf /opt/python/cp26-cp26mu /opt/_internal/cpython-2.6.9-ucs4
+RUN rm -rf /opt/python/cp33-cp33m /opt/_internal/cpython-3.3.6
+RUN rm -rf /opt/python/cp34-cp34m /opt/_internal/cpython-3.4.6
+
+FROM base as cuda
+ARG BASE_CUDA_VERSION=10.1
```
malfet, line 38: Please update it to 10.2, as 10.1 is no longer supported
```suggestion
ARG BASE_CUDA_VERSION=10.2
```

manywheel/Dockerfile_2014
```diff
@@ -0,0 +1,143 @@
...
+ADD ./common/install_cuda.sh install_cuda.sh
+RUN bash ./install_cuda.sh ${BASE_CUDA_VERSION} && rm install_cuda.sh
+
+FROM base as intel
+# MKL
+ADD ./common/install_mkl.sh install_mkl.sh
+RUN bash ./install_mkl.sh && rm install_mkl.sh
+
+FROM base as magma
+ARG BASE_CUDA_VERSION=10.1
```
malfet, line 49:
```suggestion
ARG BASE_CUDA_VERSION=10.2
```
Same as above

manywheel/Dockerfile_2014
```diff
@@ -0,0 +1,143 @@
...
+COPY --from=libpng             /usr/local/bin/png*                   /usr/local/bin/
+COPY --from=libpng             /usr/local/bin/libpng*                /usr/local/bin/
+COPY --from=libpng             /usr/local/include/png*               /usr/local/include/
+COPY --from=libpng             /usr/local/include/libpng*            /usr/local/include/
+COPY --from=libpng             /usr/local/lib/libpng*                /usr/local/lib/
+COPY --from=libpng             /usr/local/lib/pkgconfig              /usr/local/lib/pkgconfig
+COPY --from=jni                /usr/local/include/jni.h              /usr/local/include/jni.h
+
+FROM common as cpu_final
+ARG BASE_CUDA_VERSION=10.1
```
malfet, line 112:
```suggestion
ARG BASE_CUDA_VERSION=10.2
```

malfet (approved):

mattip:
I updated the BASE_CUDA_VERSION to 10.2. 

Should the update of cuda 10.1 to 10.2 also be done in the other places it appears:
```
libtorch/ubuntu16.04/Dockerfile
manywheel/Dockerfile
manywheel/build.sh
```

snnn:
Not after long you will need to move to manylinux2 because the latest CUDA release requires libstdc++ from GCC 5.x but the one you got in manylinux2014 is based on GCC 4.x. 

"CUDA Math Libraries toolchain uses C++11 features, and a C++11-compatible standard library (libstdc++ >= 20150422) is required on the host." from: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

manywheel/Dockerfile_2014
```diff
@@ -0,0 +1,143 @@
...
+
+RUN yum install -y wget curl perl util-linux xz bzip2 git patch which perl zlib-devel
+RUN yum install -y yum-utils centos-release-scl
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms
+RUN yum install -y devtoolset-7-gcc devtoolset-7-gcc-c++ devtoolset-7-gcc-gfortran devtoolset-7-binutils
+ENV PATH=/opt/rh/devtoolset-7/root/usr/bin:$PATH
+ENV LD_LIBRARY_PATH=/opt/rh/devtoolset-7/root/usr/lib64:/opt/rh/devtoolset-7/root/usr/lib:$LD_LIBRARY_PATH
+
+# cmake
+RUN yum install -y cmake3 && \
```
snnn, line 19: As you are using manylinux2014_x86_64 as the base, it should already have cmake and gcc. So do you need to install it again?

mattip: If it is already installed, yum will not install it again.

mattip:
> Not after long you will need to move to manylinux2_24 ...

... which apparently also will not be enough, see the discussion in issue #520. But this will be a first step in that direction.

mattip:
Any ideas what happened to github actions?

manywheel/Dockerfile_2014
snnn: manylinux installs cmake  from pypi. yum doesn't know that. 
And, they don't have gcc7, they installed gcc 9. So when you install GCC7, you will get two GCCs.

rgommers:
> Any ideas what happened to github actions?

That's not very useful CI output:

image

Never seen that before. Just restart it by close/open or pushing a new commit?

mattip:
@seemethere @malfet could you start CI?

mattip:
It crashed again. I rebased off master maybe that will solve it.
@seemethere @malfet please restart when you get a chance

PR #797, opened 2021-07-09
libtorch: Move libtorch images to bionic

seemethere:
Xenial is EOL, let's move to bionic

Relates to https://github.com/pytorch/pytorch/issues/61460

### DIFF
--- a/libtorch/build_docker.sh
+++ b/libtorch/build_docker.sh
@@ -16,14 +16,16 @@ case ${CUDA_VERSION} in
     ;;
 esac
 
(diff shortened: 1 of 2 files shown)


libtorch/build_docker.sh
```diff
@@ -16,14 +16,16 @@ case ${CUDA_VERSION} in
     ;;
 esac
 
+DOCKER=${DOCKER:-docker}
```
seemethere, line 19: This is to allow me to do things like this easily:

```
DOCKER=nerdctl libtorch/build_docker.sh
```

PR #810, opened 2021-07-22
Include Vulkan in nightly builds

SS-JIA:


malfet (changes requested):
Please update PR description to include the following:
 - Expected binary size increase
 - `ldd -r libtorch.so` output that demonstrates that binary does not depend on VULKAN
Also, please share link (and/or add automated tests) that demonstrates that vulkan integration function as expected and returns structure exception (one that could be caught by Python runtime), when vulkan runtime is not found

PR #818, opened 2021-07-30
aarch64 OpenBLAS: set NUM_THREADS=64 to support aarch64 instances upto 64 cores

snadampal:
This patch sets the max thread count allowed on any instance is 64.
However, the actual openblas threads used are controlled by application runtime
settings.

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -183,7 +183,7 @@ def install_condaforge(host: RemoteHost) -> None:
 def build_OpenBLAS(host: RemoteHost, git_clone_flags: str = "") -> None:
     print('Building OpenBLAS')
     host.run_cmd(f"git clone https://github.com/xianyi/OpenBLAS -b v0.3.15 {git_clone_flags}")
-    host.run_cmd("pushd OpenBLAS; make USE_OPENMP=1 NO_SHARED=1 -j8; sudo make USE_OPENMP=1 NO_SHARED=1 install; popd")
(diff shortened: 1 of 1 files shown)


malfet (approved):
Change looks good to me, but please sign CLA first

snadampal:
Thanks, @malfet. I'm working on the CLA signing process.

nSircombe:
Hi @snadampal, @malfet,

Just as an additional check, I've been trying this out with various combinations of machine and `NUM_THREADS` and it all checks out, I can reproduce the original issue on c6g builds and get builds from 8-core t4g that work over the full 64 threads of a c6g.16xlarge.

malfet:
@nSircombe and how did you solve the TLS cannot be allocated issue?

nSircombe:
I've not been using the script from `builder` for these checks - I've been using these builds from Arm's 'Tool Solutions' repo: https://github.com/ARM-software/Tool-Solutions/tree/master/docker/pytorch-aarch64

I don't get the TLS error on these, the only time I've seen it is this recent attempt to use the script from the `builder` repo.

The Tool Solutions builds are, of course, not portable manylinux whls though.

nSircombe:
It's probably worth noting that hardwiring to 64 cores (while better than getting a build that tops out at 8) isn't ideal for silicon with >64 cores, as some N1 SoCs have.

snadampal:
I agree with you @nSircombe, as the number of cores become > 64, this parameters needs change. 
Unfortunately there is no runtime solution possible because OpenBLAS defines its internal buffer allocations based on the num_threads (more details here:  https://fossies.org/linux/OpenBLAS/USAGE.md).
One thing I was thinking earlier was to make the num_threads arbitrarily large (eg:256) to cover next few generation SoCs, but refrained from it due to unnecessary memory overhead that might come with it.

PR #830, opened 2021-08-14
Enable llvm for pip wheels

bertmaher:


bertmaher:
@seemethere @malfet I've verified that https://github.com/pytorch/pytorch/pull/63272 fixes torchaudio (I realized the fastest way to debug was to do a ci/binaries build with that diff and ssh into the builder container to fiddle around, and also I can download the binaries from there and test against torchaudio to *ensure* I've got a canonically-built package).

Anyways, if you can please merge this (and #831; I separated pip and conda so they could be independently reverted if necessary) at your earliest convenience, we can give it another go.

malfet (approved):

PR #832, opened 2021-08-17
add cuda install logs

mszhanyi:
Fix https://github.com/pytorch/pytorch/issues/63139

### DIFF
--- a/windows/internal/cuda_install.bat
+++ b/windows/internal/cuda_install.bat
@@ -191,7 +191,7 @@ if %ERRORLEVEL% NEQ 0 (
     echo There's another installer running.
 )
 
-start /wait setup.exe -s %ARGS%
+start /wait setup.exe -s %ARGS% -loglevel:6 -log:"%cd%/cuda_install_logs"
 
 tasklist /fi "imagename eq setup.exe" | find ":"
 if %ERRORLEVEL% NEQ 0 (
--- a/windows/internal/cuda_install.bat
+++ b/windows/internal/cuda_install.bat
@@ -185,12 +185,6 @@ echo Installing CUDA toolkit...
 7z x %CUDA_SETUP_FILE% -o"%SRC_DIR%\temp_build\cuda"
 pushd "%SRC_DIR%\temp_build\cuda"
 
-rem Why find ":", if no process, the output is INFO: No tasks are running which match the specified criteria.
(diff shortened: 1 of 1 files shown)


mszhanyi:
@peterjc123 @seemethere 
The verification PR is
https://github.com/pytorch/pytorch/pull/62874

windows/internal/cuda_install.bat
```diff
@@ -185,22 +185,15 @@ echo Installing CUDA toolkit...
...
 pushd "%SRC_DIR%\temp_build\cuda"
 
-rem Why find ":", if no process, the output is INFO: No tasks are running which match the specified criteria.
-tasklist /fi "imagename eq msiexec.exe" | find ":"
-if %ERRORLEVEL% NEQ 0 (
-    echo There's another installer running.
-)
-
-start /wait setup.exe -s %ARGS%
+sc query wususerv
```
peterjc123, line 188: typo? also check line 190

windows/internal/cuda_install.bat
```diff
@@ -185,22 +185,15 @@ echo Installing CUDA toolkit...
...
+sc config wuauserv start= disabled
```
peterjc123, line 189: cc @seemethere @malfet I'm not sure whether it will break something at the CircleCI side or not. Would you please open a ticket about it?

windows/internal/cuda_install.bat
mszhanyi: I think both `sc query wususerv` could be removed

windows/internal/cuda_install.bat
mszhanyi: Windows update would start again after `net stop wususerv`

peterjc123: But you set it to disabled instead of stopping it here, right?

windows/internal/cuda_install.bat
peterjc123: Yeah, it's only used for debugging purposes

windows/internal/cuda_install.bat
mszhanyi: Yes. So far, CI only pass as the update service is disabled.

seemethere: I feel like ideally they wouldn’t be re-using runners between runs so this shouldn’t be a big issue.

malfet (approved):

PR #835, opened 2021-08-24
[Win] Call randomtemp from launcher wrapper script

peterbell10:
Tested in pytorch/pytorch#63782

This updates the windows builds to use `randomtemp.exe` as a compiler launcher, instead of trying to make it spoof nvcc. Since `CUDA_NVCC_EXECUTABLE` has to be a single file, I write `randomtemp.exe sccache.exe nvcc.exe [remaining args]` into a batch file, and pass that to CMake. 

This is needed for pytorch/pytorch#62445 since some CMake versions need `CMAKE_CUDA_COMPILER` to be the real nvcc, in order to find the CUDA toolkit directory.

cc @malfet


malfet (approved):

malfet:
LGTM, update PR title a bit

PR #837, opened 2021-08-24
[libtorch] Populate /opt/python in docker images

peterbell10:
Loooking at this failing CI job:
https://app.circleci.com/pipelines/github/pytorch/pytorch/368932/workflows/1f95e9a9-04d4-4d30-866c-95c60bd62cc7/jobs/15575031

I see that even though `DESIRED_PYTHON` is 3.7m, the actual python on path is 3.9 from conda. This happens with all libtorch nightly builds. The build scripts add `/opt/python/$DESIRED_PYTHON` to the path, but the directory doesn't exist so conda's python is used instead.

cc @malfet

### DIFF
--- a/libtorch/Dockerfile
+++ b/libtorch/Dockerfile
@@ -17,6 +17,11 @@ FROM base as openssl
 ADD ./common/install_openssl.sh install_openssl.sh
 RUN bash ./install_openssl.sh && rm install_openssl.sh
 
+# Install python
+FROM base as python
+COPY manywheel/build_scripts /build_scripts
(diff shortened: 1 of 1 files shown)


malfet (approved):

malfet:
Thank you very much for fixing that. Although I'm a bit confused, why we are attempting to build numpy as part of libtorch build in the first place

peterbell10:
pip tries to install numpy 1.15 into python 3.9, for which there are no wheels available because it's not supported. So, pip falls back to building from source.

PR #838, opened 2021-08-24
[libtorch] Populate /opt/python in docker images

peterbell10:
Follow up to #837 which didn't work.

The previous PR was missing the `COPY` directive, so the command wasn't actually being run. Once I fixed this, the docker build failed because the manylinux scripts use `yum` instead of `apt`. So, I've factored the CPython build script into `common/install_cpython.sh` and added the `apt-get` calls into the `Dockerfile`.

cc @malfet


manywheel/Dockerfile
```diff
@@ -62,10 +63,6 @@ ADD ./common/install_patchelf.sh install_patchelf.sh
 RUN bash ./install_patchelf.sh && rm install_patchelf.sh
 RUN cp $(which patchelf) /patchelf
 
-FROM base as breakpad
-ADD ./common/install_breakpad.sh install_breakpad.sh
-RUN bash ./install_breakpad.sh
```
peterbell10, line 67: This fails because it pins a commit that has been rebased out of the repo's history. Deleted because `breakpad` is now a git submodule in pytorch, since pytorch/pytorch#63186

malfet (approved):

manywheel/build_scripts/build.sh
```diff
@@ -44,8 +41,7 @@ autoconf --version
 # against a recent openssl [see env vars above], which is linked
 # statically. We delete openssl afterwards.)
 build_openssl $OPENSSL_ROOT $OPENSSL_HASH
-mkdir -p /opt/python
-build_cpythons $CPYTHON_VERSIONS
+/build_scripts/install_cpython.sh
```
malfet, line 44: Shouldn't the relative path be used here?

peterbell10: This script is only used for manywheel so it's fine.

PR #841, opened 2021-09-02
move 2 binaries to S3

### LINKED ISSUE #788: Replace dropbox and google drive links with s3 links
https://github.com/pytorch/builder/blob/71a2b9aaf0b4d56a2b6048b92ba71858f3717dbb/windows/internal/cuda_install.bat#L174-L182

We should replace these dependencies with S3 assets, @peterjc123 do you know what's hosted on these?

cc @mszhanyi

mszhanyi:
Fix #788


mszhanyi:
@seemethere @peterjc123

peterjc123 (changes requested):
Please see my comment. https://github.com/pytorch/vision/pull/4058/files#r701067987

mszhanyi:
> Please see my comment. https://github.com/pytorch/vision/pull/4058/files#r701067987

@peterjc123 , that's a different issue. Time is a little long, I don't remember very clearly why the snapshot is in that place.
I first find the version conflict issue is in https://github.com/pytorch/vision/pull/4058#issuecomment-862532742, which broke the vision build.
For Pytorch build, it might be OK because Pytorch build is in CPU machine. But vision build is in GPU machine, I think the reason is vision build & test is in one workflow.

malfet:
I think PR title is a bit misleading: It does not move 2 binaries, it moves one and then deletes another.
Please update PR description and also crate a draft PR against pytorch/pytorch repo that tests the change (similar to https://github.com/pytorch/pytorch/pull/62307 ) and mention it in the test plan,

mszhanyi:
@malfet , there's a test PR https://github.com/pytorch/pytorch/pull/63551 for it. I planned to add the verification link when all tests passed.

At first, I planned to moved 2 binaries to S3, which the commit 3a10296b502e76c158126e4a02485f95b6cc71e0 did.
Because when I was helping vision to enable cuda test on Windows. I found the gpu_driver_dlls.zip is only needed for cu102.

But vision has skipped all cuda tests in their CI now, so it's a little hard to me to verify it.
Since Pytorch build is on CPU machine, I'm trying to remove gpu_driver_dlls at all to see the result

mszhanyi:
per malfet's suggestion. This PR only update the download URL.

It'd better to merge it soon because  4 nightly CI workflows failed on yesterday (Sep 7) due to
image

image

image

the CI should exit once the zip isn't downloaded completely.
I'll add it in another PR.

PR #842, opened 2021-09-07
Build macOS pip wheel with llvm

bertmaher:
Since we have a bit of extra time before the 1.10 release and the probabilistic programming team has asked for it, I'd like to start building the macOS nightlies with cpu fuser support.  This PR adds it for pip (conda is forthcoming).


bertmaher:
@malfet @seemethere Tested in https://github.com/pytorch/pytorch/pull/64552

seemethere (approved):

PR #843, opened 2021-09-07
Build macOS conda binary with llvm

bertmaher:
Enabling llvm/cpu fusion on macOS conda builds.


bertmaher:
@malfet @seemethere tested in https://github.com/pytorch/pytorch/pull/64519

seemethere (approved):

PR #846, opened 2021-09-08
improve handling windows update

mszhanyi:
fix https://github.com/pytorch/pytorch/issues/64536

1. Add `sc stop wuauserv` as well
2. Restore windows update after cuda installation.
3. Improve readability of the log output.

testing PR https://github.com/pytorch/pytorch/pull/63551

### DIFF
--- a/windows/internal/cuda_install.bat
+++ b/windows/internal/cuda_install.bat
@@ -186,6 +186,9 @@ echo Installing CUDA toolkit...
 pushd "%SRC_DIR%\temp_build\cuda"
 
 sc config wuauserv start= disabled
+sc stop wuauserv
+sc query wuauserv
+
 start /wait setup.exe -s %ARGS% -loglevel:6 -log:"%cd%/cuda_install_logs"
 echo %errorlevel%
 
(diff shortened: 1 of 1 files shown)


mszhanyi:
@peterjc123 @malfet @seemethere

malfet:
@mszhanyi please do not re-enable wsupdate( as it's likely slow down test), otherwise looks good to me.

mszhanyi:
> @mszhanyi please do not re-enable wsupdate( as it's likely slow down test), otherwise looks good to me.

@malfet Could we re-enable them for some time?
I'm a little afraid that the reason why circleci machines start wsupdate frequently AGAIN in this week is that we have disabled the wsupdate for a long time. From Aug20 to Aug27, there isn't any errors caused by the wsupdate. Maybe we need to give these machine an chance to update.

I think what we might to do is to ensure no wsupdate when installing cuda.

seemethere (approved):

seemethere:
> > @mszhanyi please do not re-enable wsupdate( as it's likely slow down test), otherwise looks good to me.
> 
> @malfet Could we re-enable them for some time?
> I'm a little afraid that the reason why circleci machines start wsupdate frequently AGAIN in this week is that we have disabled the wsupdate for a long time. So disabling the service become invalid. From Aug20 to Aug27, there isn't any errors caused by the wsupdate. Maybe we need to give these machine an chance to update.
> 
> I think what we might to do is to ensure no wsupdate when installing cuda.

We shouldn't be afraid of this, the CircleCI machines shouldn't be long lived so we shouldn't worry about turning updates off on them

windows/internal/cuda_install.bat
```diff
@@ -232,3 +237,7 @@ echo Installing GPU driver DLLs
 
 echo Cleaning temp files
 rd /s /q "%SRC_DIR%\temp_build" || ver > nul
+
+sc config wuauserv start= auto
+sc start wuauserv
+sc query wuauserv
```
malfet, line 243: Let's not re-enable it on a short lived instances
```suggestion
```

PR #848, opened 2021-09-10
analytics: Add ability to pass remote

seemethere:
Adds the ability to pass remote ass an arg to the cli, also adds a log
message to note which remote is being used

### DIFF
--- a/analytics/github_analyze.py
+++ b/analytics/github_analyze.py
@@ -362,7 +362,7 @@ def main():
         # Pick best remote
         remote = next(iter(remotes.keys()))
         for key in remotes:
-            if remotes[key] == 'https://github.com/pytorch/pytorch':
+            if remotes[key].endswith('github.com/pytorch/pytorch'):
                 remote = key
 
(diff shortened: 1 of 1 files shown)


analytics/github_analyze.py
```diff
@@ -352,12 +356,14 @@ def parse_arguments():
...
-    for key in remotes:
-        if remotes[key].endswith('github.com/pytorch/pytorch'):
-            remote = key
+    remote = args.remote
+    if not remote:
+        remotes = get_git_remotes(args.repo_path)
+        # Pick best remote
+        remote = next(iter(remotes.keys()))
+        for key in remotes:
+            if remotes[key] == 'https://github.com/pytorch/pytorch':
```
malfet, line 365: Please use `endswith` instead of equal (as all my remotes look like `https://malfet:XXXX@github.com/pytorch/pytorch`)
```suggestion
            if remotes[key].endswith('github.com/pytorch/pytorch'):
```

malfet (approved):

PR #881, opened 2021-11-03
set git jobs number when checkout submodule

frank-dong-ms:
Use environment variable to set git jobs number when checkout submodule, reason for that is I see git stuck when checkout submodule with docker.

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -101,13 +101,9 @@ else
     pushd $pytorch_rootdir
 fi
 
-if [[ -n "$GIT_JOBS_NUMBER" ]]; then
-    git_jobs_number="$GIT_JOBS_NUMBER"
-else
-    git_jobs_number=0
-fi
+git_update_num_jobs="${GIT_UPDATE_NUM_JOBS:=0}"
 
-git submodule update --init --recursive --jobs $git_jobs_number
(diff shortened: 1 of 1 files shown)


manywheel/build_common.sh
```diff
@@ -100,7 +100,14 @@ if [[ ! -d "$pytorch_rootdir" ]]; then
 else
     pushd $pytorch_rootdir
 fi
-git submodule update --init --recursive --jobs 0
+
+if [[ -n "$GIT_JOBS_NUMBER" ]]; then
```
driazati, line 104: 1. can we rename to `GIT_UPDATE_NUM_JOBS` for clarity?
2. bash has a language feature for this pattern, could you switch to that to make this simpler? https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash

frank-dong-ms: Sure, will rename it and use the recommend pattern, thanks for pointing out.

driazati (approved):

driazati:
Code looks fine, might need a rebase though before we can merge

frank-dong-ms:
I looked at main fork, looks like submodule checkout has been moved out to circle ci, we will no longer need this PR, I will close this PR, thanks @driazati for review.

PR #884, opened 2021-11-13
Enable mpi install during build pytorch from source

frank-dong-ms:
Enable mpi install during build pytorch from source

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -34,6 +34,7 @@ if [[ "$OS_NAME" == *"CentOS Linux"* ]]; then
         source /etc/profile.d/modules.sh
         export OMPI_MCA_opal_cuda_support=true
         module load mpi
+    fi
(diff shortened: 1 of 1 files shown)


malfet (changes requested):
Can you please explain the intent of this PR? Also, can you please add a link to a dummy test PR to make sure it does not cause any regressions?

Looks like this change enables OpenMPI integration with PyTorch wheels, which would introduce runtime dependency and violate manylinux1 constraint

frank-dong-ms:
> Can you please explain the intent of this PR? Also, can you please add a link to a dummy test PR to make sure it does not cause any regressions?
> 
> Looks like this change enables OpenMPI integration with PyTorch wheels, which would introduce runtime dependency and violate manylinux1 constraint

Thanks for you suggestion, yes, install OpenMPI here doesn't make sense, will find other way to accomplish this.

PR #890, opened 2021-11-17
Removed caffe2 testing in LTS binary builds

mstfbl:
This PR carries over the changes done in PR #864, which removes the caffe2 import tests. This is causing the LTS binary builds to fail, due to it perceiving a missing NumPy package. [Sample failure link](https://app.circleci.com/pipelines/github/pytorch/pytorch/409789/workflows/9f84dfcc-1ad9-4108-9e1a-985be6f3ee0d/jobs/16726528?invite=true#step-105-2878).

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -300,7 +300,6 @@ if [[ "$PACKAGE_TYPE" == 'libtorch' ]]; then
 else
   pushd /tmp
   python -c 'import torch'
-  python -c 'from caffe2.python import core'
   popd
 fi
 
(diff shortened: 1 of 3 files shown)


malfet (changes requested):
Hmm, this looks like a functional change to me, which should not be allowed in LTS branch - Caffe2 support will be removed starting from PyTorch-1.11, but is still present in 1.10 and definitely were present at the time of 1.8 release? So why do you want to disable it?

mstfbl:
@malfet Thank you for your comment, I mistakenly assumed that PR #864 would be a change that'd need to be ported to the LTS branch. I'll close this PR.

PR #898, opened 2021-11-24
Adding nvidia docker image build

atalman:


.github/workflows/build-nvidia-images.yml
```diff
@@ -0,0 +1,51 @@
...
+  build-nvidia-docker:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Build Nvidia Docker Image
+        run: |
+          git clone https://gitlab.com/nvidia/container-images/cuda.git
+          cd cuda
+          PLATFORM_ARG=`printf '%s ' '--platform'; for var in $(echo $ARCHES | sed "s/,/ /g"); do printf 'linux/%s,' "$var"; done | sed 's/,*$//g'`
+          cp NGC-DL-CONTAINER-LICENSE dist/${CUDA_VERSION}/${OS}/base/
+          docker buildx build --load ${PLATFORM_ARG} -t "${IMAGE_NAME}:${CUDA_VERSION}-base-${OS_BASE}" "dist/${CUDA_VERSION}/${OS}/base"
```
seemethere, line 44: Should we build each of these in their own job so that they can run in parallel?

.github/workflows/build-nvidia-images.yml
```diff
@@ -0,0 +1,51 @@
...
+env:
+  BUILD_ENVIRONMENT: build-linux-magma
+  DOCKER_REGISTRY: "docker.io"
+  DOCKER_BUILDKIT: 0
+  COMPOSE_DOCKER_CLI_BUILD: 0
+  DOCKER_ID: ${{ secrets.DOCKER_ID }}
+  DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
+  IN_CI: 1
+  IS_GHA: 1
+  IMAGE_NAME: "nvcr.io/nvidia/cuda"
```
seemethere, line 28: Maybe we should instead choose to upload these to `ghcr.io`
```suggestion
  IMAGE_NAME: "ghcr.io/pytorch/nvidia/cuda"
```

.github/workflows/build-nvidia-images.yml
atalman: No, since runtime uses base and then devel uses runtime like this: base->runtime->devel

.github/workflows/build-nvidia-images.yml
atalman: Done!

seemethere (approved):

PR #900, opened 2021-12-01
Using default workers for manywheel and libtorch jobs

atalman:
Changing workers from linux.2xlarge  to ubuntu-18.04 workers. This way we wont need to to use so much resources, specially all the jobs in the builder package.

### DIFF
--- a/.github/workflows/build-libtorch-images.yml
+++ b/.github/workflows/build-libtorch-images.yml
@@ -23,10 +23,10 @@ env:
 
 jobs:
   build-docker:
-    runs-on: linux.2xlarge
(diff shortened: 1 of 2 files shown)


janeyx99 (approved):

malfet:
Can you please add some reasoning behind the change
Also, why CUDA support matrix is drastically reduced, but rocm for example left intact?

atalman:
> Can you please add some reasoning behind the change Also, why CUDA support matrix is drastically reduced, but rocm for example left intact?

As per offline conversation, CUDA support matrix restored. Only worker type have changed.

PR #904, opened 2021-12-06
manywheel: consolidate build into one dispatcher

seemethere:
Consolidates all of our build shell scripts into one dispatcher that can
determine which script to use based on GPU_ARCH_TYPE.

Original behavior should be preserved by doing the default behavior
based on whether or not GPU_ARCH_TYPE actually exists as an env variable
(it shouldn't in CircleCI)


seemethere:
Verifying this change here: https://github.com/pytorch/pytorch/pull/69490

seemethere:
Verified here: https://app.circleci.com/pipelines/github/pytorch/pytorch/420616/workflows/c7179878-cb71-40f6-b70f-c48a9e4f175a/jobs/16827487

PR #905, opened 2021-12-07
WIP Add binary builds github actions workflows

seemethere:


seemethere:
We went with a different approach

PR #931, opened 2022-01-18
Fix py3.10 check binary tests

atalman:
Fix py3.10 check binary tests
Test plan:
bash -c 'desired_python=3.9; echo ${desired_python:0:4}'
3.9
bash -c 'desired_python=3.10; echo ${desired_python:0:4}'
3.10

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -27,7 +27,7 @@ if [[ "$PACKAGE_TYPE" == libtorch ]]; then
   # NOTE: Only $PWD works on both CentOS and Ubuntu
   install_root="$PWD"
 else
-  py_dot="${DESIRED_PYTHON:0:3}"
+  py_dot="${DESIRED_PYTHON:0:4}"
   install_root="$(dirname $(which python))/../lib/python${py_dot}/site-packages/torch/"
 fi
 


janeyx99 (approved):

check_binary.sh
```diff
@@ -27,7 +27,7 @@ if [[ "$PACKAGE_TYPE" == libtorch ]]; then
   # NOTE: Only $PWD works on both CentOS and Ubuntu
   install_root="$PWD"
 else
-  py_dot="${DESIRED_PYTHON:0:3}"
+  py_dot="${DESIRED_PYTHON:0:4}"
```
janeyx99, line 30: why don't we just use DESIRED_PYTHON instead of py_dot then if the variable is unchanged?

atalman: I think we want to make sure only required number of digits get in - no minor verisons:)

seemethere: Would it be better to just do a `cut` based on using `.` as our delimeter?

Or even just set `py_dot` to `DESIRED_PYTHON` if it accurately reflects what we expect.

i.e. if `DESIRED_PYTHON` matches `[1-9]*\.[1-9]+` then we just set `py_dot=${DESIRED_PYTHON}`

seemethere:
Reverting this PR, looks like it caused downstream builds to fail: https://app.circleci.com/pipelines/github/pytorch/pytorch/442093/workflows/fc28c5c2-585d-4037-8e05-a455c92894c9/jobs/16956664

PR #939, opened 2022-01-20
[LTS] Pin delocate package to before python 3.6 drop

jambayk:
The latest version of delocate package drops support for Python 3.6. So, the package is pinned to an older commit before the drop.

### DIFF
--- a/wheel/build_wheel.sh
+++ b/wheel/build_wheel.sh
@@ -181,7 +181,7 @@ python setup.py bdist_wheel -d "$whl_tmp_dir"
 echo "Finished setup.py bdist_wheel at $(date)"
 
 echo "delocating wheel dependencies"
-retry pip install https://github.com/matthew-brett/delocate/archive/master.zip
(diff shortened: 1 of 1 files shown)


wheel/build_wheel.sh
```diff
@@ -181,7 +181,7 @@ python setup.py bdist_wheel -d "$whl_tmp_dir"
 echo "Finished setup.py bdist_wheel at $(date)"
 
 echo "delocating wheel dependencies"
-retry pip install https://github.com/matthew-brett/delocate/archive/master.zip
+retry pip install https://github.com/matthew-brett/delocate/archive/a5ba7dcc8ccfc9aaa1411c63266bfaf87a4ff393.zip
```
malfet, line 184: Use tags rather than hash sum
```suggestion
retry pip install https://github.com/matthew-brett/delocate/archive/refs/tags/0.10.2.zip
```

malfet (approved):

PR #949, opened 2022-01-27
use PYTORCH_ROOT for pytorch_rootdir

seemethere:
These values shouldn't be hardcoded to anything and should be set by
upstream workflows. This doesn't remove the original behavior though and
makes it backwards compat safe. So if no PYTORCH_ROOT is specified then
it should function exactly the same.


seemethere:
Related to failures observed in https://github.com/pytorch/pytorch/runs/4970254970?check_suite_focus=true

atalman (approved):

PR #954, opened 2022-02-03
[ROCm] Rename shared libraries to fix Kineto (roctracer runtime loading) issue

jithunnair-amd:
The Kineto feature is currently broken for PyTorch wheels. We get the following error when trying to use Kineto feature:
`roctracer: Loading 'libamdhip64.so' failed, libamdhip64.so: cannot open shared object file: No such file or directory`

Using the Kineto feature loads `libroctracer.so`, which in turn tries to dlopen `libamdhip64.so` and `libroctx64.so`. Those doesn't exist in the wheel install path, since the wheel build flow adds hashes to the library names. This PR disables hashing for ROCm, and *in addition*, renames the numbered .sos to their non-numbered versions eg. `libamdhip64.so.4`->`libamdhip64.so`.

Verified with PyTorch unit tests that use Kineto eg.
```
test_profiler.py: TestProfiler.test_export_stacks
test_linalg.py: TestLinalgCUDA.test_norm_fused_type_promotion_cuda_bfloat16
test_jit_fuser_te.py: TestTEFuser.test_profiler
test_autograd.py: TestAutograd.test_profiler
test_utils.py: TestBottleneck.test_bottleneck_cuda
```

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -316,21 +316,18 @@ for pkg in /$WHEELHOUSE_DIR/torch*linux*.whl /$LIBTORCH_HOUSE_DIR/libtorch*.zip;
                 cp $filepath $destpath
             fi
 
-            patchedpath=$(fname_with_sha256 $destpath)
+            # ROCm workaround for roctracer dlopens
+            if [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
+                patchedpath=$(fname_without_so_number $destpath)
+            else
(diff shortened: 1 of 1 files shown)


jithunnair-amd:
cc @jeffdaily

jeffdaily (approved):
LGTM

jithunnair-amd:
@seemethere Can you please review this and merge?

seemethere (approved):

PR #959, opened 2022-02-10
[ROCm] add support for libtorch packages

jeffdaily:
install_rocm.sh is copied from latest in pytorch that already had support for both CentOS and Ubuntu.  install_rocm.sh was further modified due to differences in CI image needs from pytorch versus packaging image needs here.


seemethere (approved):
LGTM, going to wait for green signal to merge

PR #972, opened 2022-02-23
Add cmake min version and pkg-config to build_aarch64_wheel script

atalman:
Add cmake min version and pkg-config to build_aarch64_wheel script 

This is to fix following issues:
1: Using the Ninja generator requires CMake version 3.13 or greater
2: Missing pkg-config


atalman:
not needed when building with docker

PR #974, opened 2022-02-24
build_aarch64_wheel:  Increase VolumeSize for ec2 instance to avoid running out of space for aarch64 builds

atalman:
Minor change, build_aarch64_wheel:  Increase VolumeSize for ec2 instance to avoid running out of space for aarch64 builds

### DIFF
--- a/build_aarch64_wheel.py
+++ b/build_aarch64_wheel.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# This script builids and test AARCH64 wheels using EC2.
+# This script is for building  AARCH64 wheels using AWS EC2 instances.
 # To generate binaries for the release follow these steps:
 # 1. Update mappings for each of the Domain Libraries by adding new row to a table like this:  "v1.11.0": ("0.11.0", "rc1"),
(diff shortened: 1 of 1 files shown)


build_aarch64_wheel.py
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env python3
 
+# This script builids and test AARCH64 wheels using EC2.
```
malfet, line 3:
```suggestion
# This script is for building  AARCH64 wheels using AWS EC2 instances.
```

build_aarch64_wheel.py
```diff
@@ -50,7 +58,16 @@ def start_instance(key_name, ami=ubuntu18_04_ami, instance_type='t4g.2xlarge'):
                                 SecurityGroups=['ssh-allworld'],
                                 KeyName=key_name,
                                 MinCount=1,
-                                MaxCount=1)[0]
+                                MaxCount=1,
+                                BlockDeviceMappings=[
```
malfet, line 62: Can we make it an option?
Also, are volumes automatically destroyed when terminate instances is called?

malfet (approved):

Issue #969, opened 2022-02-22
Potential Typo in builder


ydaiming:
Should this be `MANY_LINUX_VERSION` instead of `MANY_LINUX_VERSON`?
https://github.com/pytorch/builder/blob/3e6551ca3b205d17fa9570e747c87fdf8df1449c/manywheel/build_docker.sh#L56

qhaas:
Noticed this last night, `MANY_LINUX_VERSON` appears only once in this entire repo, almost certainly a typo and they meant `MANY_LINUX_VERSION`

PR #979, opened 2022-02-26
fix typos per issue 969

qhaas:
As described in [this comment of issue 969](https://github.com/pytorch/builder/issues/969#issuecomment-1050842719), it appears two different typos (`MANYLINUX_VERSION` vs `MANY_LINUX_VERSION` and `MANY_LINUX_VERSION` vs `MANY_LINUX_VERSON`) hid the fact the environment variable `MANYLINUX_VERSION` was being ignored and `docker.io/pytorch/manylinux-builder` is built instead of the expected `docker.io/pytorch/manylinux2014-builder` when `2014` is set.

`MANY_LINUX_VERSION` is assumed to be the intended variable name.


qhaas:
My employer (Oak Ridge National Laboratory) has reached out to `cla@fb.com` concerning the CLA.

Jmennius (approved):
Why not merge this? The bug/typo is definitely present and should be fixed.
I imagine that some scrutineering is required to make sure that what CI/CD is producing does not change unexpectedly (that is if they use `build_all_docker.sh` at all, which we don't really know).

PR #987, opened 2022-03-09
Explicitly turn off USE_CUPTI_SO

ezyang:
This is currently the default but I'm changing the default
in https://github.com/pytorch/pytorch/pull/66539 ; so this
change is necessary to preserve builder behavior.


malfet (approved):
Don't you need to do similar change to conda/build.sh? (and also add it to pass-thru environment variables in `pytorch-nightly/meta.yaml`)

ezyang:
testing at https://github.com/pytorch/pytorch/pull/74008

when I ran binary tests on conda, they didn't fail, only the wheel. So I'm guessing I only need it here to start...

ezyang:
I guess the claim I'm making is that we don't have to statically link cupti on conda. Time will tell...

PR #992, opened 2022-03-18
[WIP] ADD 11.6 workflow for docker image build

ptrblck:
* adds CUDA 11.6 docker builds
* adds GitHub workflows for 11.6

* note: I assume the `magma` install fails as we would need to build magma for 11.6 and upload it to conda first

CC @atalman

### DIFF
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: linux.2xlarge
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


atalman (approved):
I think this is fine the fact that magma fails since it depends on the built Docker file hence we need to deploy this PR first and then post magma PR

atalman:
actually @ptrblck don't add libtorch chganges for now. This should fix magma issue. So we have following steps. 1 conda/Dockerfile changes, Magma Changes, Libtorch changes

atalman (approved):

PR #1004, opened 2022-03-25
Revert "bump cudnn to 8.3.2.44 for CUDA11.3"

malfet:
Reverts pytorch/builder#1000


malfet:
Shouldn't be needed if #1005 will work

Issue #1010, opened 2022-03-31
Pin CMake to be consistent with Pytorch


zengk95:
Addressing https://github.com/pytorch/pytorch/pull/74954 followup

PR #1011, opened 2022-03-31
Pin cmake 3.22.3

zengk95:
Addresses #1010

Searched for all usages of `install.*cmake`

### DIFF
--- a/libtorch/Dockerfile
+++ b/libtorch/Dockerfile
@@ -5,7 +5,7 @@ FROM ${GPU_IMAGE} as base
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get clean && apt-get update
-RUN apt-get install -y curl locales git-all autoconf automake make cmake wget unzip
+RUN apt-get install -y curl locales git-all autoconf automake make cmake=3.22.3 wget unzip
 
 RUN locale-gen en_US.UTF-8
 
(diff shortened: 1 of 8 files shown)


malfet (approved):

manywheel/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN wget http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm &
     rm -f epel-release-latest-7.noarch.rpm
 
 # cmake
-RUN yum install -y cmake3 && \
+RUN yum install -y cmake3-3.22.3 && \
```
malfet, line 23: Same as above
```suggestion
RUN yum install -y cmake3 && \
```

libtorch/Dockerfile
```diff
@@ -42,7 +42,7 @@ ENV CUDA_HOME /usr/local/cuda
 FROM base as conda
 ADD ./common/install_conda.sh install_conda.sh
 RUN bash ./install_conda.sh && rm install_conda.sh
-RUN /opt/conda/bin/conda install -y cmake=3.14
+RUN /opt/conda/bin/conda install -y cmake=3.22.3
```
malfet, line 45: Let's stick with 3.14 as it allows us to validate that CMakeFile is correct at minspec
```suggestion
RUN /opt/conda/bin/conda install -y cmake=3.14
```

manywheel/conda_build.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-yum install -y wget git cmake
+yum install -y wget git cmake-3.22.3
```
malfet, line 4:
```suggestion
yum install -y wget git cmake
```

malfet (changes requested):
Do not pin version in `apt-get`/`yum install` command, as cmake=3.23 is not available there yet (which is probably reflected in CI failures)

zengk95:
@malfet I removed those pins as suggested. Could you review again?

malfet (approved):
Thank you for the updates, looks good to me

malfet:
Looks like it caused MacOS night jobs to fail, as cmake-3.22.3 is not available from Anaconda, is it: https://anaconda.org/anaconda/cmake

PR #1021, opened 2022-04-19
[ROCm] build libdrm from source for custom amdgpu.ids path

jeffdaily:
libdrm hard-codes the path to the file amdgpu.ids.  Packaging ROCm dependencies into the wheel relocates the libdrm library making the original path invalid, resulting in an extra warning when using the wheels.

/opt/amdgpu/share/libdrm/amdgpu.ids: No such file or directory

Adding the file amdgpu.ids to the auxiliary file list and building libdrm from source to replace the hard-coded path with a dynamic path resolves the issue.

### DIFF
--- a/common/install_rocm_drm.sh
+++ b/common/install_rocm_drm.sh
@@ -83,7 +83,7 @@ index a5007ffc..a3627529 100644
 +	if (count > 0) {
 +		self_path[count] = '\0';
 +
-+		// replace bin/python with lib
++		// remove '/bin/python' from self_path
 +		for (i=count; i>0; --i) {
 +			if (self_path[i] == '/') break;
 +			self_path[i] = '\0';
@@ -93,7 +93,7 @@ index a5007ffc..a3627529 100644
 +			if (self_path[i] == '/') break;
 +			self_path[i] = '\0';
 +		}
-+		strcat(self_path, "lib");
++		self_path[i] = '\0';
 +
 +		if (0 == nftw(self_path, check_for_location_of_amdgpuids, 5, FTW_PHYS)) {
 +			if (amdgpuids_path) {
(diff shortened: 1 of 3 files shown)


jeffdaily:
@seemethere I don't have authority to rerun the failed builds.  They only failed due to a lack of available GHA runners.

common/install_rocm_drm.sh
```diff
@@ -0,0 +1,132 @@
...
++
+ 	/* 1st valid line is file version */
+ 	while ((n = getline(&line, &len, fp)) != -1) {
+ 		/* trim trailing newline */
+EOF
+
+###########################
+### build
+###########################
+meson builddir --prefix=/opt/amdgpu
```
Mystro256, line 127: You can trim this down a bit if you want, as ROCm doesn't need the full libdrm. ROCm doesn't need intel support, or nvidia (nouveau) support, or manpages, or andreno (freedreno)

E.g.

meson -Dintel=false -Dnouveau=false -Dfreedreno=false -Dcairo-tests=false -Dman-pages=false builddir --prefix=/opt/amdgpu

jeffdaily: I doubt the unnecessary additions bloat the final library size much.  Would it be okay to leave as-is since the CI already passed for the current settings?

jeffdaily:
Ready to merge?

seemethere (approved):

common/install_rocm_drm.sh
Mystro256: It's probably fine as-is, but it would be good to follow up to optimise. While the size is likely not a huge deal, I'm not sure what the functionality impact is, as with ROCm, we don't build the full libdrm, so this is technically an untested superset of libdrm.

PR #1043, opened 2022-06-01
Manywheel: mark all git directories as safe

malfet:
`gitcore-2.36+` refuses to run commands in repos not owned by the user
invoking the command. As `pytorch` and `builder` folders are mounted
into the image, set `safe.directory` global parameter to "*"

Fixes https://github.com/pytorch/pytorch/issues/78659


atalman (approved):
LGTM

PR #1044, opened 2022-06-01
binary: Add check for git_version

seemethere:
Should catch regressions like https://github.com/pytorch/pytorch/issues/78659

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -316,6 +316,7 @@ fi
 if [[ "$PACKAGE_TYPE" != 'libtorch' ]]; then
   pushd /tmp
   python -c 'import torch; assert torch.git_version != "Unknown"'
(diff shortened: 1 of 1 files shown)


check_binary.sh
```diff
@@ -310,6 +310,15 @@ else
   popd
 fi
 
+###############################################################################
+# Check torch.git_version
+###############################################################################
+if [[ "$PACKAGE_TYPE" != 'libtorch' ]]; then
+  pushd /tmp
+  python -c 'import torch; assert torch.git_version != "Unknown"'
```
atalman, line 318: We should also check for None here

atalman (approved):
LGTM

check_binary.sh
seemethere: Added!

malfet (approved):

PR #1056, opened 2022-06-10
Fix libtorch static builds that regressed

powderluv:
libtorch-static used to ship libtorch.a until 1.0.1 and has
regressed since forcing people to build from source.

This fixes the build commands and add BUILD_SHARED_LIBS=OFF
It also removes build it as a debug build, and turns off
USE_TENSORPIPE until https://github.com/pytorch/tensorpipe/issues/449
is fixed.

Tested on my system to verify a libtorch.a is produced but can't
test the CI

Fixes: https://github.com/pytorch/pytorch/issues/70898

PR #1063, opened 2022-06-16
Revert "Fix libtorch static builds that regressed"

malfet:
Reverts pytorch/builder#1056 as it breaks static builds, see https://hud.pytorch.org/pytorch/pytorch/commit/a4fa62064e1863d108c655f541bce4d806cf1823

### DIFF
--- a/manywheel/build_libtorch.sh
+++ b/manywheel/build_libtorch.sh
@@ -122,14 +122,8 @@ fi
 
 echo "Calling setup.py install at $(date)"
 
-BUILD_SHARED_VAR="ON"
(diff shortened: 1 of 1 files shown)


powderluv:
I have isolated the breakage to USE_MKL / USE_MKLDNN generating wrong Ninja rules like ` $<TARGET_FILE:dnnl_graph> ` 

I have filed https://github.com/pytorch/pytorch/issues/80012 to fix the issue in PyTorch but will land a fix for static builds with MKL disabled for now until it is fixed.

PR #1065, opened 2022-06-21
[Win] Update cuda-11.6.0 to 11.6.2 (#1064)

atalman:
According to https://github.com/pytorch/pytorch/issues/79583#issuecomment-1161270425 its cusparse_static does not try to mess with global locale


malfet (approved):

malfet:
I don't know if `cusprase_static` even exists on Windows nor that we use it, but its better to be consistent across the platforms

Issue #1067, opened 2022-06-27
libtorch static builds run out of builder system memory


powderluv:
We re-enabled true static builds 

But it fails with error 137: 

https://github.com/pytorch/pytorch/runs/7059088258?check_suite_focus=true

Error 137 says:

https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory#:~:text=This%20is%20a%20Linux%20error,memory%2C%20your%20build%20will%20fail.

which indicates the linking OOMs the docker container.  Should we disable tests for static builds ?

powderluv:
@malfet FYI.  Is there a way to increase system RAM ? Or should I send a PR to disable tests on static builds ?

Willian-Zhang:
Bump this, we need this. 

BTW, I was not able to find a guide for static building until landing on this issue. I am still jumping between PR code changes and CD pipelines.

powderluv:
Unless someone has internal CI access I don't think we can land this as an external PR.

Willian-Zhang:
> Unless someone has internal CI access I don't think we can land this as an external PR.

and we might need circle ci team on this also. ([related thread](https://discuss.circleci.com/t/increase-4g-memory-limit/9070/5))
Just curious, do you know if PyTorch is using an paid plan?

PR #1068, opened 2022-06-27
Disable TESTS for static library builds

powderluv:
This fixes issue https://github.com/pytorch/builder/issues/1067

This reduces the system memory requirements but ideally we want
beefier containers to run the tests too but this can be a temp fix

### DIFF
--- a/manywheel/build_libtorch.sh
+++ b/manywheel/build_libtorch.sh
@@ -123,6 +123,7 @@ fi
 echo "Calling setup.py install at $(date)"
 
 BUILD_SHARED_VAR="ON"
+BUILD_TESTS_VAR="ON"
 USE_TENSORPIPE_VAR="ON"
 USE_MKLDNN_VAR="ON"
 if [[ $LIBTORCH_VARIANT = *"static"* ]]; then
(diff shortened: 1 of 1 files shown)


powderluv:
@malfet here is a workaround for static builds to reduce the link time requirements on the CI bot.

malfet (approved):

malfet:
Sure, though I wonder if we shoudl instead just kill static libtorch builds

powderluv:
oh no oh no. we really want the static builds downstream for projects like `torch-mlir` so we don't accidentally link against another version of libtorch | pytorch installed in the python paths (.dylibs and such).

malfet:
Reverting, as it does not fix build, see https://github.com/pytorch/pytorch/runs/7082027627?check_suite_focus=true for example.

Can you please test changes you make against builder, by proposing a draft PR against pytorch that targets a different builder branch, for example see https://github.com/pytorch/pytorch/pull/74141 (feel free to modify the dispatch logic of binary builds to run on PRs if you can not attach `ciflow/binary_wheels` label)

As for linker OOMs: have you tried using gold linker?

powderluv:
Thank you for the pointer on how to test the builder changes. In the sample PR I think external contributors will not be able to run unless we have our patch on a branch on pytorch/builder rather than a per user fork of the builder repo. e.g: my patch here was on [powderluv:disable-tests-static](https://github.com/powderluv/builder/tree/disable-tests-static) and I am not sure how we can make .github/templates/common.yml.j2 pick up a branch from a fork.

makslevental:
@powderluv pretty sure the hack is to patch here as well https://github.com/pytorch/pytorch/blob/6bce0f4acecc8d9d0b790299eb39c1adca59209b/.github/templates/macos_binary_build_workflow.yml.j2#L86

```
repository="pytorch/builder"...
```

PR #1072, opened 2022-06-27
Add scripts to do pypi binary promotion

seemethere:
Adds scripts to promote pypi binaries to a staging environment as well
as promote them again to the production environment


release/pypi/promote_pypi_to_production.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+
+# Make sure to update these versions when doing a release first
+PYTORCH_VERSION=${PYTORCH_VERSION:-1.11.0}
```
malfet, line 6: Do you want to update those to 1.12 already?

malfet (approved):

release/pypi/promote_pypi_to_production.sh
seemethere: Yup, should be updated in the latest

PR #1079, opened 2022-07-01
Add MAGMA patches fixing LU factorization

IvanYashchuk:
These patches should `fix` https://github.com/pytorch/pytorch/issues/70111 and https://github.com/pytorch/pytorch/issues/57482. However, the failure in these issues is observed only on Windows CI (I think it's GPU dependent, not OS) and this package is only for Linux. I saw these failures on Linux on P100 (`getrf_shfl.patch`) and V100 (`getrf_nbparam.patch`) GPUs.

What would be the procedure for updating the MAGMA lib that's used in Windows CI?

### DIFF
--- a/magma/package_files/meta.yaml
+++ b/magma/package_files/meta.yaml
@@ -11,7 +11,7 @@ source:
      - getrf_nbparam.patch  # fixes "too many resources requested for launch" error
 
 build:
-  number: 0
+  number: 1
   script_env:
     - DESIRED_CUDA
     - CUDA_ARCH_LIST
--- a/magma/package_files/meta.yaml
+++ b/magma/package_files/meta.yaml
@@ -7,6 +7,8 @@ source:
    patches:
      - cmakelists.patch
      - thread_queue.patch
+     - getrf_shfl.patch  # incorrect results for <7.0 arch
+     - getrf_nbparam.patch  # fixes "too many resources requested for launch" error
 
 build:
   number: 0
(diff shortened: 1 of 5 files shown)


atalman (approved):

magma/package_files/meta.yaml
```diff
@@ -7,6 +7,8 @@ source:
    patches:
      - cmakelists.patch
      - thread_queue.patch
+     - getrf_shfl.patch  # incorrect results for <7.0 arch
+     - getrf_nbparam.patch  # fixes "too many resources requested for launch" error
 
 build:
   number: 0
```
IvanYashchuk, line 14: This has to be bumped before merging.

IvanYashchuk: Done.

IvanYashchuk:
Rebuild for Linux will be triggered automatically after the merge, right?
https://github.com/pytorch/builder/blob/main/.github/workflows/build-magma-linux.yml

atalman:
> Rebuild for Linux will be triggered automatically after the merge, right? https://github.com/pytorch/builder/blob/main/.github/workflows/build-magma-linux.yml

 Does not look like it will, could you please add
```
magma/package_files/*
```

To path: in both places
 https://github.com/pytorch/builder/blob/main/.github/workflows/build-magma-linux.yml#L10
and  
https://github.com/pytorch/builder/blob/main/.github/workflows/build-magma-linux.yml#L15

we should then see the workflows running in this PR before merging, and then merge on green

IvanYashchuk:
@atalman, could you please approve running workflows?

PR #1082, opened 2022-07-07
Pass fabi version 11 to all gcc builds

atalman:
Pass fabi version 11 to all gcc builds


malfet (approved):
LGTM but please add the test and create a PyTorch PR to verify that it works)

atalman:
not needed

PR #1085, opened 2022-07-12
fix syntax error thats causing red on pytorch

clee2000:
see https://github.com/pytorch/pytorch/runs/7304042943?check_suite_focus=true


janeyx99 (approved):
Yes, but we shouldn't have landed the change that intro'd this

PR #1096, opened 2022-08-04
[WIP] s3_management: add a mechanism to manage third-party dependencies cache on S3

izaitsevfb:
Partially addresses: pytorch/pytorch#75703

This PR is currently a work in progress, see below for the list of TODOS.

# Problem

As described in pytorch/pytorch#75703 currently there are multiple places where build is dependent on third-party dependencies hosted on on third-party servers (which shown to be **not reliable**).

Some of these places are just basic `wget` and `curl` requests of third-party URLs, that can (and should be) cached in our public S3 buckets to provide a fallback for reliability.

The issue with manual uploading of such dependencies is two-fold, first, it's time consuming, second, external engineers don't have access to `ossci` S3 buckets.

The issue with fully automated upload is the lack of transparency, control, and security.

# Proposal

This PR proposes a compromise between manual and fully automated caching of third-party dependencies. 

The script (see `s3_management/thirdparty_deps/manage.py`) that takes a `yml` configuration with the list of urls with corresponding S3 keys:
```yml
- bucket: ossci-linux
  url: https://developer.download.nvidia.com/compute/redist/cudnn/v8.3.2/local_installers/11.5/cudnn-linux-x86_64-8.3.2.44_cuda11.5-archive.tar.xz
  key: cudnn/cudnn-linux-x86_64-8.3.2.44_cuda11.5-archive.tar.xz 

- bucket: ossci-linux
  url: https://developer.download.nvidia.com/...
  key: cuda/...

...
```
...and uploads URLs (that haven't been synced yet) to S3, making them public.

The script is intended to be invoked by CI (that has access to `ossci` S3 secrets).

### Intended workflow

* The yml config will be committed to one (or perhaps more) of pytorch org repos
    * the corresponding CI workflows will be added as well
* Sync script is invoked by CI when config is pushed into `master` branch, or manually
    * This way PyTorch Dev Infra team can validate and approve the changes in the S3 cache
    * Adding new URLs to the S3 cache and using them elsewhere could be done within the same PR (may require manual triggering of the sync workflow)

### Validation

Script contains rudimentary auto validation (currently hardcoded) of the yml config:
   * white list of bucket names
   * list of allowed key prefixes for each bucket
   * regex s3 key validation
   * url validation

But the workflow mostly relies on manual review of the PRs containing the S3 cache update.

# Advantages
* Self-documenting organization of the `ossci` S3 buckets (at least the third-party cache part)
   * yml config in git allows to trace the context of who and why uploaded a file to S3
* Sharing the ability to upload to S3 with external engineers in a controlled way
* Uploading to s3 is automated -> saves time

# TODO
- [ ] Add CI jobs that invoke the script
   - [ ] discuss which projects should include the yml config and the CI job (builder, pytorch?)
- [ ] Extract the validation config into the yml file
- [ ] Add documentation and reference from other parts of the documentation that require S3 upload
- [ ] Add more validation rules (?)

# Potential issues/questions
* potential race condition when sync jobs with two versions of the conf with the same key are run in parallel
   * could happen:
       * when one sync job is triggered manually
       * when two PRs that **update** the same key are merged in the rapid succession
   * undefined outcome: either file version will be on S3
* updating/removing dependencies is not supported automatically
   * currently manual deletion from S3 is required


izaitsevfb:
The next step is to add the CI job(s) that will run the script to perform the sync according to the config.

And it's not clear for me what is the best project to place the workflow and sync config in.

The options are:
* `pytorch/builder` project
* `pytorch/pytorch` project
* both projects: have a separate config per project

Since there are cached dependencies in both project, it might be reasonable to have separate configs in each project, at the same time it's easier to maintain a config that is in one place.

Note: `pytorch/pytorch` already contains another s3-related workflow [update_s3_htmls.yml](https://github.com/pytorch/pytorch/blob/master/.github/workflows/update_s3_htmls.yml).

@janeyx99 @huydhn @atalman , since you already know the context, could you please share you thoughts on that?

huydhn:
> potential race condition when sync jobs with two versions of the conf with the same key are run in parallel

I think we could avoid this by not running the script at pull_request or push, but as a periodical job. We probably don't really need to have sub-second sync here, do we? It also keeps things simple :)

> pytorch/builder

My understand is that this repo is dedicated to building and releasing pytorch. So I feel that having the CI job here is a bit out of place.

> pytorch/pytorch

We could have it here but I rather not do this because merging a PR into this repo takes a long time (without test target determination and other cool things we are planning to build ;)). IMO, we don't really need this such extensive testing capability for this change.

Thus the best place I can think of is https://github.com/pytorch/test-infra. As its name indicates, it is a collection of infrastructure components that are supporting the PyTorch CI/CD system.

huydhn:
FYI, I have another prime candidate for this script besides CUDA stuffs, https://github.com/pytorch/pytorch/blob/master/.jenkins/pytorch/win-test-helpers/installation-helpers/install_miniconda3.bat#L12. This is definitely is going to be useful :)

izaitsevfb:
After the discussion with @seemethere, I decided to stick with the dockerised GH action approach, as it allows to have granular config per project and is more reliable (when image is built it is guaranteed to work).

The system will consist of two parts:
1. in `pytorch/builder` there will be a docker image of the GH action with the sync script (this PR). The image must be built and pushed to docker repo once and it won't change often (if ever).
2. in other projects there will be a configuration + workflow to apply the configuration
    * See an example of the new workflow and configuration in this PR: pytorch/pytorch#83306
    * Similar workflows and configs could be added to other projects (beside `pytorch/pytorch`) as needed to cache files on S3 that are specific to these projects.

I believe this PR is ready for review.

s3_management/thirdparty_deps/manage.py
```diff
@@ -0,0 +1,110 @@
...
+            print(f'Invalid key format: {key}')
+            exit(1)
+
+        allowed_prefixes = allowed_buckets_with_key_prefixes.get(bucket)
+        if allowed_prefixes is None:
+            print(f'Bucket {bucket} is not allowed. Here is the list of allowed buckets: '
+                  f'{list(allowed_buckets_with_key_prefixes.keys())}')
+            exit(1)
+
+        if not any(key.startswith(prefix) for prefix in allowed_prefixes):
```
huydhn, line 98: Minor: Is this possible to bypass this check completely with a key like `miniconda/../dir-that-i-not-suppose-to-touch/Miniconda3-latest-Windows-x86_64.exe`.  Although this is used by us, so we probably don't want to do this for ourselves ;)

s3_management/thirdparty_deps/manage.py
```diff
@@ -0,0 +1,110 @@
...
+            print(f'Invalid key: {key}, allowed key prefixes for bucket {bucket}: {allowed_prefixes}')
+            exit(1)
+
+        if exists_on_s3(bucket, key):
```
huydhn, line 102: Hmm, does this work for file like Miniconda3-latest-Windows-x86_64.exe in which there might be a newer version of the same name?  IMO, having something like aws s3 sync https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html would be best.  That command does several things:

1. Send HEAD request to S3 to get the file metadata
2. If the file exists, compare the size and time stamp with the local file (IIRC). There are options `--size-only` and `--exact-timestamps` to tune this.  Basically, the idea is that if the file doesn't change, no need to do anything
3. If the file does change, i.e. different size in bytes, then uploading the local file

s3_management/thirdparty_deps/manage.py
```diff
@@ -0,0 +1,110 @@
...
+import yaml
+
+boto_config = Config(
+  retries={
+    'max_attempts': 3,
+    'mode': 'standard'
+  }
+)
+
+S3 = boto3.resource('s3')
```
huydhn, line 20: Is this variable used anywhere?  I see only the client is being used

s3_management/thirdparty_deps/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.8
```
huydhn, line 1: For this, could you double check to see if `3.8-alpine` works?  The difference in size is worth the try, python:3.8 size is 861MB while 3.8-alpine is only 46MB.  I don't think we need any complicated things here besides base python and pip

s3_management/thirdparty_deps/manage.py
izaitsevfb: Great idea!

The problem with this approach as is would be that downloading the file is still required to calculate its checksum.

However, I think I can adapt this idea in to following way: have the ability to (optionally) specify the file checksum manually in the config, and, if doesn't match the one on S3, the file will be re-uploaded.

s3_management/thirdparty_deps/manage.py
izaitsevfb: Awesome spot!

s3_management/thirdparty_deps/Dockerfile
izaitsevfb: Probably makes sense to check https://github.com/pytorch/builder/blob/main/s3_management/Dockerfile as well.

izaitsevfb: On the other hand, it may not matter, because the base image is popular and probably already cached on github runners, so only the diff will be actually pulled.

izaitsevfb: Changed base to  `python:3.8-alpine` and verified that is works.

s3_management/thirdparty_deps/manage.py
izaitsevfb: Disabled `/../` in the key validation.

s3_management/thirdparty_deps/manage.py
izaitsevfb: I think I'll defer this suggestion (I've created an issue pytorch/builder#1103 to track it).

The rationale:
* Having the mechanism to update files is certainly nice to have, but it's not essential
* I want to land the initial version ASAP and start using it for real-world scenarios to see how it fares
* I want to minimize the scope of this PR

@huydhn, what do you think?

huydhn: Sounds good. We can get this one done first. It would still work for us as it stands right now. I'm happy to track this issue separately

huydhn (approved):

PR #1104, opened 2022-08-15
[aarch64] build pytorch wheel with mkldnn and acl backend

snadampal:


malfet (commented):
Hmm, can you split this PR into few smaller ones and provide a more detailed description about some of the proposed changes.

build_aarch64_wheel.py
```diff
@@ -407,12 +411,13 @@ def configure_system(host: RemoteHost, *,
     if not use_conda:
         host.run_cmd("sudo apt-get install -y python3-dev python3-yaml python3-setuptools python3-wheel python3-pip")
     host.run_cmd("pip3 install dataclasses typing-extensions")
-    # Install and switch to gcc-8 on Ubuntu-18.04
-    if not host.using_docker() and host.ami == ubuntu18_04_ami and compiler == 'gcc-8':
-        host.run_cmd("sudo apt-get install -y g++-8 gfortran-8")
-        host.run_cmd("sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 100")
-        host.run_cmd("sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-8 100")
-        host.run_cmd("sudo update-alternatives --install /usr/bin/gfortran gfortran /usr/bin/gfortran-8 100")
```
malfet, line 415: Sorry, but why are you deleting this codepath? Please provide an explanation, why supporting ubuntu-18.04 is wrong here?

build_aarch64_wheel.py
```diff
@@ -602,7 +608,7 @@ def parse_arguments():
     parser.add_argument("--debug", action="store_true")
     parser.add_argument("--build-only", action="store_true")
     parser.add_argument("--test-only", type=str)
-    parser.add_argument("--os", type=str, choices=list(os_amis.keys()), default='ubuntu18_04')
+    parser.add_argument("--os", type=str, choices=list(os_amis.keys()), default='ubuntu20_04')
```
malfet, line 611: Why does default needs to be changed from 18.04 to 20.04?

malfet (changes requested):
See comments and please provide a test plan for the PR. (and for example share a link on the wheel build by this script)
I.e. one of the challenges I had in the past, it to build a wheel that works on both Ubuntu and RedHat OSes. Perhaps it would be good to add a test step that could be invoked for wheels already available on calleers local machine)

build_aarch64_wheel.py
snadampal: Hi @malfet, thanks for the review! We need gcc-10 for Arm Compute Library multi-arch support, that's I replaced gcc-8 with gcc-10. But as you suggested, I will split this into incremental commits with more context.

malfet: I think increasing min supported version to gcc-10 requires some more indepth discussion. I.e. is it compatible with aarch64 manylinux2014 requirements as outlined in https://peps.python.org/pep-0599/

snadampal:
@malfet ,updated the pr with multiple commits addressing the latest tools required for pytorch build and enabling mkldnn+acl backend.

build_aarch64_wheel.py
```diff
@@ -18,11 +18,10 @@
 
 # AMI images for us-east-1, change the following based on your ~/.aws/config
 os_amis = {
-    'ubuntu18_04': "ami-0f2b111fdc1647918",  # login_name: ubuntu
     'ubuntu20_04': "ami-0ea142bd244023692",  # login_name: ubuntu
```
malfet, line 21: Not sure why are you removing ubutnu18-04
```suggestion
    'ubuntu18_04': "ami-0f2b111fdc1647918",  # login_name: ubuntu
    'ubuntu20_04': "ami-0ea142bd244023692",  # login_name: ubuntu
```

build_aarch64_wheel.py
```diff
@@ -18,11 +18,10 @@
...
     'redhat8': "ami-0698b90665a2ddcf1",  # login_name: ec2-user
 }
-ubuntu18_04_ami = os_amis['ubuntu18_04']
+ubuntu20_04_ami = os_amis['ubuntu20_04']
```
malfet, line 24:
```suggestion
ubuntu18_04_ami = os_amis['ubuntu18_04']
ubuntu20_04_ami = os_amis['ubuntu20_04']
```

build_aarch64_wheel.py
```diff
@@ -189,6 +188,8 @@ def update_apt_repo(host: RemoteHost) -> None:
     host.run_cmd("sudo systemctl stop unattended-upgrades.service || true")
     host.run_cmd("while systemctl is-active --quiet apt-daily.service; do sleep 1; done")
     host.run_cmd("while systemctl is-active --quiet unattended-upgrades.service; do sleep 1; done")
+    host.run_cmd("wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null")
+    host.run_cmd("sudo add-apt-repository 'deb https://apt.kitware.com/ubuntu/ focal main'")
```
malfet, line 192: Why this change is necessary? If you want a newer cmake, let's install int from pip/conda

build_aarch64_wheel.py
```diff
@@ -214,15 +215,19 @@ def install_condaforge_python(host: RemoteHost, python_version="3.8") -> None:
     else:
         install_condaforge(host)
         # Pytorch-1.10 or older are not compatible with setuptools=59.6 or newer
-        host.run_cmd(f"conda install -y python={python_version} numpy pyyaml setuptools=59.5.0")
-
+        host.run_cmd(f"conda install -y python={python_version} numpy pyyaml setuptools=59.5.0; python3 --version")
```
malfet, line 218: Why this is necessary. And please keep 2 lines separator between functions/classes but one line between class methods

build_aarch64_wheel.py
```diff
@@ -300,7 +305,9 @@ def build_torchvision(host: RemoteHost, *,
     if host.using_docker():
         build_vars += " CMAKE_SHARED_LINKER_FLAGS=-Wl,-z,max-page-size=0x10000"
 
-    host.run_cmd(f"cd vision; {build_vars} python3 setup.py bdist_wheel")
+    # LD_PRELOAD the libtorch_cpu.so to work around the 'cannot allocate memory in static TLS block' issue
+    host.run_cmd(f"export LD_PRELOAD=$HOME/miniforge3/lib/python{python_version}/site-packages/torch/lib/libtorch_cpu.so; cd vision; {build_vars} python3 setup.py bdist_wheel")
```
malfet, line 309: This is not a valid workaround, we need to figure out what is causing it

malfet (changes requested):
LD_PRELOAD the libtorch_cpu.so to work around the 'cannot allocate memory in static TLS block' issue is not a valid workaround, and huge usability regression from the previous iteration

build_aarch64_wheel.py
snadampal: Hi @malfet, even for the manylinux2014 docker based builds, we are still dependent on the host OS for gfortran workaround in the script at:
https://github.com/pytorch/builder/blob/main/build_aarch64_wheel.py#L444

and to support ubuntu 18.04 host build (one of the supported options in this script), it needs lot of tool updates which probably not worth going forward. 

> pytorch library builds require python >=3.7 and cmake >=3.13. The default
> versions of python and cmake on ubuntu 18.04 are much older. Given ubuntu 18.04
> EOL is April 2023, the effort to update the script for ubutnu 18.04 changes
> is not worth, hence deprecating ubuntu 18.04 host os support.

build_aarch64_wheel.py
snadampal: yeah, I added this for newer cmake. Let me check conda/pip based install. thanks!

build_aarch64_wheel.py
snadampal: I did some more digging today and looks like it's OpenBLAS source built libraries consuming more TLS memory. Not sure why it wasn't observed earlier.

snadampal:
Hi @malfet, I've limited this PR only for docker (manylinux 2014) build option, so that we don't need to carry lot of tool dependencies for the host build. This is very clean now.
Here are the commands for building wheels with and without mkldnn backend

with mkldnn+acl:
`./build_aarch64_wheel.py  --python-version 3.8 --use-docker --keep-running --os ubuntu20_04 --enable-mkldnn --branch release/1.13`

without mkldnn backend (only openblas based):
`./build_aarch64_wheel.py  --python-version 3.8 --use-docker --keep-running --os ubuntu20_04 --branch release/1.13`

One pending issue I'm still debugging is...
  embedding the external libraries (Arm compute libraries in this case) into the torch wheel. I've used the embed_library.py present in this wheel builder script. The libraries are copied with additional hash tags and hence import is failing. Please let me know if you had seen this before and have any ideas....

snadampal:
Hi @malfet , I'm done with my changes. Built the image in manyLinux2014 docker mode and tested it on AWS Graviton instances. Please let me know if there are comments.

with mkldnn+acl:
./build_aarch64_wheel.py --python-version 3.8 --use-docker --keep-running --os ubuntu20_04 --enable-mkldnn --branch release/1.13

without mkldnn backend (only openblas based):
./build_aarch64_wheel.py --python-version 3.8 --use-docker --keep-running --os ubuntu20_04 --branch release/1.13

malfet (approved):

PR #1105, opened 2022-08-15
release: Make macos platform a variable / glob

seemethere:
macos binaries changed between the last promotion to have a different
version string, change these to be a variable with a glob so we can just
search for macos + ARCH to make this more resilient

Relates to https://github.com/pytorch/audio/issues/2620


atalman (approved):
LGTM

PR #1101, opened 2022-08-09
[WIP] Testing release validation automation

atalman:

PR #1107, opened 2022-08-18
Smoke test binary

atalman:
Tests: https://github.com/pytorch/builder/pull/1101
Usage:
```
conda run -p ${ENV_NAME} python3  ./test/smoke_test/smoke_test.py
```

### DIFF
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -18,9 +18,9 @@ import torchvision.transforms
 cuda_version_expected = os.environ['CUDA_VER']
 is_cuda_system = cuda_version_expected != "cpu"
 #todo add torch, torchvision and torchaudio tests
-print(torch.__version__)
-print(torchvision.__version__)
-print(torchaudio.__version__)
+print(f"torch: {torch.__version__}")
+print(f"torchvision: {torchvision.__version__}")
+print(f"torchaudio: {torchaudio.__version__}")
 print('Is torchvision useable?', all(x is not None for x in [torch.ops.image.decode_png, torch.ops.torchvision.roi_align]))
 if(not torch.cuda.is_available() and is_cuda_system):
     print(f"Expected CUDA {cuda_version_expected}. However CUDA is not loaded.")
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -1,3 +1,4 @@
+import os
 import torch
 import torchvision
 import torchaudio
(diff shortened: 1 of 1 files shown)


test/smoke_test/smoke_test.py
```diff
@@ -0,0 +1,35 @@
...
+import torchaudio.utils  # noqa: F401
+from torchaudio.io import StreamReader
+import torchvision.datasets as dset
+import torchvision.transforms
+cuda_version_expected = os.environ['CUDA_VER']
+is_cuda_system = cuda_version_expected != "cpu"
+#todo add torch, torchvision and torchaudio tests
+print(torch.__version__)
+print(torchvision.__version__)
+print(torchaudio.__version__)
```
izaitsevfb, line 23: minor: perhaps adding a label here would improve clarity of the output, currently it's confusion as to which version corresponds to what:
```
1.13.0.dev20220819
0.14.0.dev20220819
0.13.0.dev20220819
Is torchvision useable? True
11.6
8302
```

e.g.:
```python
print(f"torchaudio: {torchaudio.__version__}")
```

test/smoke_test/smoke_test.py
```diff
@@ -0,0 +1,35 @@
...
+print('Is torchvision useable?', all(x is not None for x in [torch.ops.image.decode_png, torch.ops.torchvision.roi_align]))
+if(not torch.cuda.is_available() and is_cuda_system):
+    print(f"Expected CUDA {cuda_version_expected}. However CUDA is not loaded.")
+    sys.exit(1)
+if(torch.cuda.is_available()):
```
izaitsevfb, line 28: nit: outer parentheses are not necessary in `if`

izaitsevfb (approved):

test/smoke_test/smoke_test.py
atalman: I think its better to use parentheses in all if statements for consistency

izaitsevfb: I checked other python files and they use the style without parentheses, but, since we don't enforce a common style in the builder project (yet?), I think it's ok, we can always reformat later, if we have the need.

atalman:
> Some post merge feedback: imports should not be at the top of the file, but rather conditional (and guarded by appropriate try catch) Otherwise, you would not get a signal for say pytorch or torchvision version if torchaudio could not get imported
> 
> I.e. isolation in https://github.com/pytorch/pytorch.github.io/blob/e3b544439fa84f28d6de79cb18cc972423f5610c/scripts/test_install.py#L76 is deliberate, as importing affects global state of the python runtime

@malfet I specifically added these as breaking changes, so we would break even if 1 module is broken. Since the target of this test is to test the exactly install command as get started page.

PR #1102, opened 2022-08-10
Testing MacOS release Validation Automation

weiwangmeta:
On a ARM based MacOS, download miniconda, create conda env, install pytorch, and do import pytorch.

PR #1115, opened 2022-08-29
[WIP] Release smoke test validation windows

izaitsevfb:
This is a first quick and dirty solution, in the venue of #1101 and #1102.

The next step towards the final solution would be to create a reusable cross-platform composite action that performs the smoke test based on the input parameters, and reuse it here and for the start page generation smoke test (see https://github.com/pytorch/pytorch.github.io/blob/464bc2f6d9492339c69a4bd1ad03d6cc1573a651/.github/workflows/check-quickstartmodule.yml).

See pytorch/pytorch#83519 and pytorch/pytorch#82725 for the context.


izaitsevfb:
Incorporated into #1122

PR #1116, opened 2022-08-30
Modify smoke tests for binary validation

atalman:
Modify smoke tests for binary validation

### DIFF
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -1,35 +1,47 @@
-import os
-import sys
 import torch
 import torchvision
 import torchaudio
-import torch.nn as nn
-import torchaudio.compliance.kaldi  # noqa: F401
-import torchaudio.datasets  # noqa: F401
-import torchaudio.functional  # noqa: F401
-import torchaudio.models  # noqa: F401
-import torchaudio.pipelines  # noqa: F401
-import torchaudio.sox_effects  # noqa: F401
-import torchaudio.transforms  # noqa: F401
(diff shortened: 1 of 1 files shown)


test/smoke_test/smoke_test.py
```diff
@@ -3,33 +3,47 @@
...
-if(not torch.cuda.is_available() and is_cuda_system):
-    print(f"Expected CUDA {cuda_version_expected}. However CUDA is not loaded.")
-    sys.exit(1)
-if(torch.cuda.is_available()):
-    if(torch.version.cuda != cuda_version_expected):
-        print(f"Wrong CUDA version. Loaded: {torch.version.cuda} Expected: {cuda_version_expected}")
+
+def smoke_test_cuda() -> None:
+    gpu_arch_ver = os.environ['GPU_ARCH_VER']
+    desired_cuda = os.environ['DESIRED_CUDA']
```
seemethere, line 9:
```suggestion
    gpu_arch_ver = os.environ.get('GPU_ARCH_VER', "")
    gpu_arch_type = os.environ.get('GPU_ARCH_TYPE', "cpu")
```

We should try to get away from DESIRED_CUDA, it's a legacy variable since we support more than just CUDA now

seemethere (commented):
We should get rid of desired_cuda

test/smoke_test/smoke_test.py
atalman: Done!

seemethere (approved):

PR #1117, opened 2022-08-31
Linux validation binaries (conda, wheel, libtorch)

atalman:
Linux binaries validation download miniconda, create conda env, install pytorch, and do import pytorch. 
Similar to: https://github.com/pytorch/builder/pull/1102

### DIFF
--- a/.github/workflows/validate-linux-binaries.yml
+++ b/.github/workflows/validate-linux-binaries.yml
@@ -44,7 +44,7 @@ jobs:
         run: |
           nvidia-smi
       - name: Instlla Conda
-        - uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: ${{ matrix.python_version }}
           auto-update-conda: true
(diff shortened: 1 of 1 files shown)


malfet:
Lots of CUDA failures seems to stem from a the fact, that you are failing to pass `--pre` argument and therefore install binaries released to pypi rather than nightly builds. 
And the fact that you are passing 10.2 and cpu tests means that you do not check the version of the installed binaries.

atalman:
> hat you are failing to pass `--pre` argument and therefore install binaries released to pypi rather than nightly builds.
> And the fact that you are passing 10.2 and cpu tests means that y

Thank you @malfet will fix it!

seemethere (approved):
Nice!

weiwangmeta (approved):
looks great!

malfet:
Why you've merged the job with so many failures? Were they all legitimate?

atalman:
@malfet Rerunning the workflow, only 3.11 wheels are failing, which are expected:
https://github.com/pytorch/builder/runs/8207224963?check_suite_focus=true

We will have to add an ignore for 3.11 python builds for now or something like yhis.
As you can see the build PR has the same issue now: https://github.com/pytorch/test-infra/pull/640
We probably want to exclude all 3.11 builds for now in matrix.

PR #1122, opened 2022-09-06
Validate windows binaries workflow

atalman:
This implements validation of windows binaries
Please note: I had to use powershell, since GPU runner does not have WSL installed and can't use bash.
Failures on the CI are valid failures

### DIFF
--- a/.github/workflows/validate-windows-binaries.yml
+++ b/.github/workflows/validate-windows-binaries.yml
@@ -1,0 +1,80 @@
+name: Validate binary images
+
+on:
+  push:
+    branches:
+      main
+    paths:
+      - .github/workflows/validate-windows-binaries.yml
+  pull_request:
+    paths:
+      - .github/workflows/validate-windows-binaries.yml
+jobs:
+  generate-conda-matrix:
+    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-windows-binaries.yml
```diff
@@ -0,0 +1,80 @@
...
+        ${{ fromJson(needs.generate-libtorch-matrix.outputs.matrix) }}
+      fail-fast: false
+    runs-on: "windows-2019"
+    env:
+      PYTHON_VERSION: ${{ matrix.python_version }}
+    steps:
+      - name: Install pytorch and smoke test
+        shell: powershell
+        run: |
+          $install = '${{ matrix.installation }}'
```
izaitsevfb, line 78: Nit: I think $install can be inlined as:
```
Invoke-WebRequest -Uri ${{ matrix.installation }} -OutFile 'libtorch.zip'
```

izaitsevfb (approved):
LGTM

.github/workflows/validate-windows-binaries.yml
atalman: Yes, I think this should work:
`
Invoke-WebRequest -Uri '${{ matrix.installation }}' -OutFile 'libtorch.zip'
`

PR #1124, opened 2022-09-07
Modifying smoke test to add more advanced validation as requested

atalman:
Modifying smoke test to add more advanced validation as requested
Fixes: https://github.com/pytorch/pytorch/issues/83519

Please note: .github/workflows/validate-linux-binaries.yml changes shall be reverted prior to commiting
Images in this PR are taken from torchvision repo

PyTorch smoke tests validate that:
torch module can be imported
3x3 convolution works on available devices (CPU/GPU)

TorchVision smoke tests validate that:
torchvision module can be imported
torchvision can be used to decode jpeg and png image
resnet50 can classify image of a dog as a dog


malfet (changes requested):
For all test, validate that decoding works (i.e. returned values are not None and have expected properties)
Do not assume that test is executed as `python foo.py`, and use `Path(__file__).parent` to get path to the folder this test is located

Also, why do you need to change the builder checkout to `atalman/builder`?

malfet:
Also, please be consistent with using single vs double quote in the file

.github/workflows/validate-linux-binaries.yml
```diff
@@ -36,8 +39,11 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.validation_runner }}
     steps:
-      - name: Checkout PyTorch builder
+      - name: Checkout PyTorch atalman/builder
         uses: actions/checkout@v2
+        with:
+          ref: main
+          repository: atalman/builder
```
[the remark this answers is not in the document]
atalman: This is temp change, otherwise the checkout/builder would use the code from builder/main that does not have the changes to the test

test/smoke_test/smoke_test.py
```diff
@@ -21,10 +21,63 @@ def smoke_test_cuda() -> None:
...
+        conv = nn.Conv2d(3, 3, 3).cuda()
+        x = torch.randn(1, 3, 24, 24).cuda()
+        with torch.cuda.amp.autocast():
+            out = conv(x)
+        print(out.sum())
+
 def smoke_test_torchvision() -> None:
     import torchvision.datasets as dset
     import torchvision.transforms
+    from torchvision.io import read_file, decode_jpeg, decode_png
```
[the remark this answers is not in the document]
atalman: Good point! Let me refactor

.github/workflows/validate-linux-binaries.yml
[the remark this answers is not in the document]
malfet: > This is temp change, otherwise the checkout/builder would use the code from builder/main that does not have the changes to the test

Either this workflow is written in a horribly broken way, or above statement is wrong; checkout action conventionally works for pull requests without any changes

test/smoke_test/smoke_test.py
```diff
@@ -21,10 +21,63 @@ def smoke_test_cuda() -> None:
...
+    m = nn.Conv2d(16, 33, (3, 5), stride=(2, 1), padding=(4, 2), dilation=(3, 1))
+    input = torch.randn(20, 16, 50, 100)
+    output = m(input)
+    if(is_cuda_system):
+        print(f"Testing smoke_test_conv2d with cuda")
+        conv = nn.Conv2d(3, 3, 3).cuda()
+        x = torch.randn(1, 3, 24, 24).cuda()
+        with torch.cuda.amp.autocast():
+            out = conv(x)
+        print(out.sum())
```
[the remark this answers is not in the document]
atalman: Hm, out return different values all the time. Can you suggest a way to modify the test so that conv2d returns const ?

atalman:
> For all test, validate that decoding works (i.e. returned values are not None and have expected properties) Do not assume that test is executed as `python foo.py`, and use `Path(__file__).parent` to get path to the folder this test is located
> 
> Also, why do you need to change the builder checkout to `atalman/builder`?

Done removed these changes.

test/smoke_test/smoke_test.py
```diff
@@ -3,29 +3,78 @@
 import torch
 import torchvision
 import torchaudio
+import pathlib
```
malfet, line 6: Please define SCRIPT_DIR global variable and use it to resolve path to object relative to script
```suggestion
from pathlib import Path

SCRIPT_DIR = Path(__file__).parent
```

test/smoke_test/smoke_test.py
```diff
@@ -3,29 +3,78 @@
...
+            out = conv(x)
 
 def smoke_test_torchvision() -> None:
-    import torchvision.datasets as dset
-    import torchvision.transforms
     print('Is torchvision useable?', all(x is not None for x in [torch.ops.image.decode_png, torch.ops.torchvision.roi_align]))
 
+def smoke_test_torchvision_read_decode() -> None:
+    from torchvision.io import read_image
+    img_jpg = read_image(str(pathlib.Path(__file__).parent / 'assets' / 'rgb_pytorch.jpg'))
```
malfet, line 45: No need to path `Path` object into string (as python is a duck-typed language)
```suggestion
    img_jpg = read_image(SCRIPT_DIR / 'assets' / 'rgb_pytorch.jpg')
```

test/smoke_test/smoke_test.py
```diff
@@ -3,29 +3,78 @@
...
+    assert img_jpg.ndim == 3 and img_jpg.numel() > 100
+    img_png = read_image(str(pathlib.Path(__file__).parent  / 'assets' / 'rgb_pytorch.png'))
```
malfet, line 47: Same as above
```suggestion
    img_png = read_image(SCRIPT_DIR  / 'assets' / 'rgb_pytorch.png')
```

.github/workflows/validate-linux-binaries.yml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
     paths:
       - .github/workflows/validate-linux-binaries.yml
+      - .test/smoke_test/smoke_test.py
```
malfet, line 12: I guess if you change the image you want to re-run this workflow as well?
```suggestion
      - .test/smoke_test/*
```

test/smoke_test/smoke_test.py
```diff
@@ -3,29 +3,78 @@
...
 
 def smoke_test_torchvision() -> None:
-    import torchvision.datasets as dset
-    import torchvision.transforms
     print('Is torchvision useable?', all(x is not None for x in [torch.ops.image.decode_png, torch.ops.torchvision.roi_align]))
 
+def smoke_test_torchvision_read_decode() -> None:
+    from torchvision.io import read_image
+    img_jpg = read_image(str(pathlib.Path(__file__).parent / 'assets' / 'rgb_pytorch.jpg'))
+    assert img_jpg.ndim == 3 and img_jpg.numel() > 100
```
malfet, line 46: Nit: (Python runtime can be configured to ignore asserts, but test is supposed to always raise an exception)
```suggestion
    if img_jpg.ndim != 3 or img_jpg.numel() < 100:
      raise RuntimeError(f'Unexpected shape of img_rgb: {img_rgb.shape}')
```

malfet (approved):
LGTM, though left a few nits.

Also, let's stick with https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings style on strings - i.e. use double quotes by default

PR #1129, opened 2022-09-08
Refactor macos and linux to use same validate-binary action

atalman:
Refactor linux and macos binaries validation


seemethere (approved):

.github/workflows/validate-linux-binaries.yml
```diff
@@ -36,10 +36,8 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.validation_runner }}
     steps:
-      - name: Checkout PyTorch builder
-        uses: actions/checkout@v2
       - name: Validate binary conda
-        uses: ./.github/actions/validate-linux-binary
+        uses: pytorch/builder/.github/actions/validate-binary@main
```
seemethere, line 40: Are you worried that using a not local version of this action will make it harder to roll out new changes?

atalman: As per discussion will move to local action rather then global one!

PR #1138, opened 2022-09-15
Test cuda of torchvision

atalman:


fduwjj:
Looks like TorchVision's cuda version does not match with that of PyTorch?

PR #1144, opened 2022-09-21
Setup periodic test to run binary verification

izaitsevfb:
fix pytorch/pytorch#84764

Make binary validation workflows reusable, parameterized with the channel and os.

### First implementation:

1. Created reusable workflow, [see  validate-binaries.yml](https://github.com/pytorch/builder/blob/2c25be546e8948975e09ce0833b91d0202d872cf/.github/workflows/validate-binaries.yml)

    The workflow (with its children) takes around **2 hours** to finish (e.g. [1h 42m 12s](https://github.com/pytorch/builder/actions/runs/3102160422))
    
    Can be invoked from pytorch/builder as
    ```jobs:
      validate-nightly-binaries:
        uses: ./.github/workflows/validate-binaries.yml
        with:
          channel: nightly
          os: all
    ```
    Or from any org repo (when merged):
    ```
    jobs:
      validate-nightly-binaries:
        uses: pytorch/builder/.github/workflows/validate-binaries.yml@main
        with:
          channel: nightly
          os: all
    ```
    Supported parameter values:
    ```
          channel:
            description: "Channel to use (nightly, test, release, all)"
            required: true
            type: choice
            options:
              - release
              - nightly
              - test
              - all
          os:
            description: "Operating system to generate for (linux, windows, macos, macos-arm64)"
            required: true
            type: choice
            options:
              - windows
              - linux
              - macos
              - all
    ```

3. Added scheduled workflow for nightly and release builds (see `validate-release-binaries.yml` and `validate-nightly-binaries.yml`) 
    * nightly channel schedule: At 2:30 pm UTC (7:30 am PDT)
    * release schedule: At 3 am and 2 pm UTC (7 am and 8 pm PDT)

----

### Next up:
* Find a way to trigger smoke on specific events (i.e. after `vision` and `audio` builds are finished)
   * Can be done via webhook trigger
   * Or as a reusable action (in this case HUD needs to track several repos)

### DIFF
--- a/.github/workflows/validate-linux-binaries.yml
+++ b/.github/workflows/validate-linux-binaries.yml
@@ -1,42 +1,49 @@
 name: Validate linux binaries
 
 on:
-  push:
-    branches:
-      main
-    paths:
-      - .github/workflows/validate-linux-binaries.yml
-      - .test/smoke_test/*
-  pull_request:
-    paths:
-      - .github/workflows/validate-linux-binaries.yml
-      - .test/smoke_test/*
+  workflow_call:
+    inputs:
+      channel:
+        description: "Channel to use (nightly, test, release, all)"
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      channel:
+        description: "Channel to use (nightly, test, release, all)"
+        required: true
+        type: choice
+        options:
+          - release
+          - nightly
+          - test
+          - all
+
 jobs:
-  generate-conda-matrix:
+  generate-linux-conda-matrix:
     uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
     with:
       package-type: conda
       os: linux
-      channel: nightly
-  generate-wheel-matrix:
+      channel: ${{ inputs.channel }}
+  generate-linux-wheel-matrix:
     uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
     with:
       package-type: wheel
       os: linux
-      channel: nightly
-  generate-libtorch-matrix:
+      channel: ${{ inputs.channel }}
+  generate-linux-libtorch-matrix:
     uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
     with:
       package-type: libtorch
       os: linux
-      channel: nightly
-
+      channel: ${{ inputs.channel }}
 
   validate-linux-binaries-conda:
-    needs: generate-conda-matrix
+    needs: generate-linux-conda-matrix
     strategy:
       matrix:
-        ${{ fromJson(needs.generate-conda-matrix.outputs.matrix) }}
+        ${{ fromJson(needs.generate-linux-conda-matrix.outputs.matrix) }}
       fail-fast: false
     runs-on: ${{ matrix.validation_runner }}
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/validate-linux-binaries.yml
```diff
@@ -1,42 +1,49 @@
...
-  push:
-    branches:
-      main
-    paths:
-      - .github/workflows/validate-linux-binaries.yml
-      - .test/smoke_test/*
-  pull_request:
-    paths:
-      - .github/workflows/validate-linux-binaries.yml
-      - .test/smoke_test/*
```
weiwangmeta, line 13: Whenever these files get changed, it would be important to trigger a run still?

.github/workflows/validate-macos-binaries.yml
```diff
@@ -1,41 +1,55 @@
 name: Validate MacOS Binaries
 
 on:
-  pull_request:
-    paths:
-      - .github/workflows/validate-macos-binaries.yml
-      - .test/smoke_test/*
```
weiwangmeta, line 7: ditto

.github/workflows/validate-nightly-binaries.yml
```diff
@@ -0,0 +1,31 @@
+name: Validate nightly binaries (all OS)
+
+on:
+  schedule:
+    # At 2:30 pm UTC (7:30 am PDT)
```
weiwangmeta, line 5: Would be good to add a comment about rough estimation of when these would complete, i.e. how long do these test take to finish.

.github/workflows/validate-windows-binaries.yml
```diff
@@ -1,40 +1,49 @@
...
-  push:
-    branches:
-      main
-    paths:
-      - .github/workflows/validate-windows-binaries.yml
-      - .test/smoke_test/*
-  pull_request:
-    paths:
-      - .github/workflows/validate-windows-binaries.yml
-      - .test/smoke_test/*
```
weiwangmeta, line 13: ditto

weiwangmeta (approved):
Looks good! Just some nit/minor suggestions.

atalman (approved):
Thank you looks Great! Please address comments from Wei!

.github/workflows/validate-linux-binaries.yml
izaitsevfb: I moved `path` trigger into `validate-nightly-binaries.yml` and `validate-release-binaries.yml`. The reason for that is that after my change OS-specific workflows require a mandatory `channel` input, and GH, as far as I can tell, doesn't support default value for the inputs when workflow is triggered on push of pull request.

Anyway, currently, workflows for both channels and all OSs will trigger if `smoke_test` changes.

.github/workflows/validate-nightly-binaries.yml
izaitsevfb: Do you think it would be sufficient to add the estimates to the PR description (since the workflow contains the link to the PR). I just don’t want to trigger the jobs for such a minor change.

izaitsevfb: Updated PR description with approximate timing: 1:40-2 hours and the link to an example run.

.github/workflows/validate-windows-binaries.yml
izaitsevfb: See comment above.

PR #1145, opened 2022-09-22
Add cuda and date check to smoke test

weiwangmeta:
Fixes https://github.com/pytorch/pytorch/issues/85085 partially. 
Addressing No.1 and No.2 in https://github.com/pytorch/pytorch/issues/85085#issuecomment-1254011982 

Essentially, this PR helps enhance the cudnn check, e.g. during development it prompted torchaudio's addition of https://github.com/pytorch/audio/pull/2707 since it was found that initially "import torchvision" would trigger cudnn check but torchaudio would not. 

Also for nightly tests, added date check to detect if (py)torch, torchvision, torchaudio are not from the same date, or if they are, whether they are obsolete (more than 2 days old).


atalman (approved):
LGTM. Probably cleaner would be to pass inputs.chanell rather then extracting 'nightly' from INSTALLATION var. But overall looks good

PR #1148, opened 2022-09-22
Trivial: fix the typo in the the job name for the release binaries validation workflow

izaitsevfb:
Currently in the HUD the jobs from nightly and release channels both have "nightly" in their name.

The issue was introduced in #1144 (it doesn't affect anything functionally, but creates a bit of confusion in the [HUD view](https://hud.pytorch.org/hud/pytorch/builder/main/1?per_page=1)).

### DIFF
--- a/.github/workflows/validate-release-binaries.yml
+++ b/.github/workflows/validate-release-binaries.yml
@@ -24,7 +24,7 @@ on:
       - .test/smoke_test/*
 
 jobs:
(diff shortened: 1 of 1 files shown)


izaitsevfb:
Cancelled the job runs to avoid compounding capacity issues.

atalman (approved):
LGTM

PR #1164, opened 2022-10-13
Regenerates RECORD file to reflect version changes in release pypi binary

syed-ahmed:
When the version in METADATA is changed (e.g. from 1.13.0+cu117.with.pypi.cudnn to 1.13.0), the hash for METADATA in the RECORD file should be changed as well. However, the binary produced before and after the script has the same hash for METADATA (e.g. `sha256=NEV-OiKmMQO5aC89LwYMAFsAedKLsWLxfZ3XdnUmRcM,23902` for the METADATA file in 1.13.0+cu117.with.pypi.cudnn binary). 

This PR copies the RECORD file regeneration logic from manywheel/build_common.sh. With this change, in the above example, the hash for METADATA is correctly regenerated in the RECORD file:
`torch-1.13.0.dist-info/METADATA,sha256=2A235SVt_S_U8gSRuhwR7fJlM5j64pMddcoYP_WmrNs,23880`

It looks like pip install doesn't care about this but the pypi spec does want installer to look at this RECORD and fail installation if things don't match: https://github.com/pypa/pip/issues/2752

cc: @malfet @atalman @ptrblck @seemethere

### DIFF
--- a/release/pypi/prep_binary_for_pypi.sh
+++ b/release/pypi/prep_binary_for_pypi.sh
@@ -12,6 +12,19 @@
 set -eou pipefail
 shopt -s globstar
 
+# Function copied from manywheel/build_common.sh
+make_wheel_record() {
+    FPATH=$1
+    if echo $FPATH | grep RECORD >/dev/null 2>&1; then
+        # if the RECORD file, then
+        echo "$FPATH,,"
+    else
+        HASH=$(openssl dgst -sha256 -binary $FPATH | openssl base64 | sed -e 's/+/-/g' | sed -e 's/\//_/g' | sed -e 's/=//g')
+        FSIZE=$(ls -nl $FPATH | awk '{print $5}')
+        echo "$FPATH,sha256=$HASH,$FSIZE"
+    fi
+}
+
 OUTPUT_DIR=${OUTPUT_DIR:-$(pwd)}
 
 tmp_dir="$(mktemp -d)"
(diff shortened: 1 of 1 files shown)


atalman (commented):
This looks good. Can you please include some testing information ? Like how the change was validated.

atalman (approved):

syed-ahmed:
I tested this as follows:

Before this PR, I just downloaded: https://download.pytorch.org/whl/test/cu117/torch-1.13.0%2Bcu117-cp38-cp38-linux_x86_64.whl and ran `prep_binary_for_pypi.sh` on it. I noticed that the version in `METADATA` is changed from `1.13.0+cu117.with.pypi.cudnn` to `1.13.0` ; however the hash for `METADATA` in the RECORD file before and after the script are the same: `sha256=NEV-OiKmMQO5aC89LwYMAFsAedKLsWLxfZ3XdnUmRcM,23902` where 23902 is the file size. Clearly if we removed some characters from the file, the file size should be shorter.

So then, I re-ran the script with this patch on the same wheel and the hash and file size in RECORD is correctly shown as: `torch-1.13.0.dist-info/METADATA,sha256=2A235SVt_S_U8gSRuhwR7fJlM5j64pMddcoYP_WmrNs,23880`

It seems like the correct way to test this thing is to use `wheel pack` and `wheel unpack`. Let me message you in slack.

atalman:
This looks good tested by @atalman

PR #1168, opened 2022-10-20
Add pypi cudnn package to tests

atalman:
Add pypi cudnn package to tests

### DIFF
--- a/.github/actions/validate-binary/action.yml
+++ b/.github/actions/validate-binary/action.yml
@@ -93,10 +93,10 @@ runs:
           set -ex
 
           # Special case Pypi installation instructions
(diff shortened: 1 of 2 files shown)


weiwangmeta (approved):
Looks good to me. Curious to know what error would we see without these changes.

atalman:
> Looks good to me. Curious to know what error would we see without these changes.

These are special builds that uploaded here: https://download.pytorch.org/whl/test/cu117_pypi_cudnn/torch/
They are normally almost the same as https://download.pytorch.org/whl/test/cu117/torch/ 
The difference is that https://download.pytorch.org/whl/test/cu117_pypi_cudnn/torch/ don't package cudnn and cublas but link to them via pypi. And yes we do need to validate these builds as well.

PR #1171, opened 2022-10-25
Quick fix to not use sccache if it's not setup properly

huydhn:
Fix binary build on HUD:

Before: https://github.com/pytorch/pytorch/actions/runs/3318218597/attempts/1
After: https://github.com/pytorch/pytorch/actions/runs/3318218597

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -4,6 +4,60 @@
 set -ex
 SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
+# Courtesy of pytorch/.jenkins/pytorch/common_utils.sh
+#
+# - 1st arg:  code to add
+# - remaining args:  names of traps to modify
+#
+trap_add() {
+    trap_add_cmd=$1; shift || fatal "${FUNCNAME[0]} usage error"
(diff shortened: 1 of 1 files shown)


malfet:
Can we please seek reviews for a non-trivial changes? And/or add line or two about test plan

huydhn:
> Can we please seek reviews for a non-trivial changes? And/or add line or two about test plan

This was a quick fix just to get HUD and nightly green.  It has been reverted now that a new docker image has been published

huydhn:
> > This was a quick fix just to get HUD and nightly green. It has been reverted now that a new docker image has been published
> 
> Sorry, what is "quick fix to get HUD green"? It would be good to say something like:
> 
> * Before this PR this job (url-to-the job) fails
> * After it passes (link)

Update the description with the link before and after.  

> Also, we should not be using sccache for binary builds at all to avoid potential corruption/stale object.

Hmm, I see that we are using this for Windows https://github.com/pytorch/builder/blob/main/windows/build_all.bat#L58 and Mac binary builds, so any reason it couldn't be used for Linux.  Note that sccache is only used in the CI, the plan is to leave nightly builds the same (not using sccache).  There is a bit more context in https://github.com/pytorch/pytorch/pull/87340 about this

PR #1176, opened 2022-10-25
Test ResNet on MPS

malfet:
After https://github.com/pytorch/pytorch/issues/86954 is fixed, we should be able to test resnet on MPS


malfet:
Mac release failures are expected, as it was broken in 1.12, but hopefully should work in 1.13

atalman (approved):
LGTM

PR #1183, opened 2022-10-28
Fixes cuda pypi rpaths and libnvrtc name

syed-ahmed:
https://github.com/pytorch/builder/pull/1181 doesn't force rpaths. Also it didn't add nvrtc rpath to the .so's in the torch/ folder.

Additionally, the libnvrtc name is changed, otherwise libcaffe2_nvrtc.so will not find it.

### DIFF
--- a/release/pypi/prep_binary_for_pypi.sh
+++ b/release/pypi/prep_binary_for_pypi.sh
@@ -61,8 +61,18 @@ for whl_file in "$@"; do
             sed -i -e "/^Requires-Dist: nvidia-cublas-cu11 (==11.10.3.66).*/a Requires-Dist: nvidia-cuda-nvrtc-cu11 (==11.7.99)" "${whl_dir}"/*/METADATA
 
             sed -i -e "s/-with-pypi-cudnn//g" "${whl_dir}/torch/version.py"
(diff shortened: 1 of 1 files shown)


atalman:
I think this is required fix:

Without this fix:
```
 patchelf --print-rpath libtorch_cuda.so
$ORIGIN/../../nvidia/cublas/lib:$ORIGIN/../../nvidia/cudnn/lib:$ORIGIN
```

With this fix:
```
patchelf --print-rpath libtorch_cuda.so
$ORIGIN/../../nvidia/cublas/lib:$ORIGIN/../../nvidia/cudnn/lib:$ORIGIN/../../nvidia/cuda_nvrtc/lib:$ORIGIN
```

malfet:
>  Also it didn't add nvrtc rpath to the .so's in the torch/ folder.
But why does it needs to? `_C.*.so` does not need to know anything about CUDA dependencies, just load files from `lib`

syed-ahmed:
When doing ldd on _C.*.so I saw dependencies on libcudnn and libcublas. Since libcudnn is not in `lib` the rpath will be needed for _C.*.so as well.

PR #1184, opened 2022-10-31
Adds infra to use NCCL from pypi

syed-ahmed:
This PR installs nccl from NVIDIA developer website, links PyTorch against this system NCCL, and points rpath to the pypi nccl wheel.

Testing plan: use this PR to build pytorch and generate a wheel. Test that wheel with some distributed workload that uses NCCL.

### DIFF
--- a/manywheel/build_cuda.sh
+++ b/manywheel/build_cuda.sh
@@ -275,15 +275,18 @@ elif [[ $CUDA_VERSION == "11.7" ]]; then
             "libcublasLt.so.11"
         )
     else
-        echo "Using cudnn and cublas from pypi."
+        echo "Using cudnn, cublas, and nccl from pypi."
         CUDA_RPATHS=(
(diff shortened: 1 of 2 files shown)


manywheel/build_cuda.sh
```diff
@@ -275,15 +275,18 @@ elif [[ $CUDA_VERSION == "11.7" ]]; then
             "libcublasLt.so.11"
         )
     else
-        echo "Using cudnn and cublas from pypi."
+        echo "Using cudnn, cublas, and nccl from pypi."
         CUDA_RPATHS=(
```
atalman, line 279: We should also add libnvrtc here

atalman (changes requested):
Looks good. Please see comment

syed-ahmed:
Reopened here https://github.com/pytorch/builder/pull/1196

PR #1194, opened 2022-11-16
remove CUDA 10.2-11.5 builds

ptrblck:
* removed CUDA 10.2-11.5 builds where possible
* might want to revisit some `Dockerfile`s which use the CUDA10.2 base image as their default, uninstall the CUDA toolkit, and reinstall the desired one afterwards

CC @atalman


atalman:
Looks like build-manywheel-images.yml, build-magma-linux.yml still have some old configs I think once these removed we should be good to go cc @ptrblck

atalman (approved):
LGTM

PR #1197, opened 2022-11-21
[Validation] Pass ref:main to general worker

atalman:
Pass ref:main to general worker

### DIFF
--- a/.github/workflows/validate-binaries.yml
+++ b/.github/workflows/validate-binaries.yml
@@ -17,6 +17,10 @@ on:
         description: "Channel to use (nightly, test, release, all)"
         required: true
         type: string
+      ref:
+        description: 'Reference to checkout, defaults to empty'
+        default: ""
+        type: string
   workflow_dispatch:
     inputs:
       os:
@@ -39,6 +43,10 @@ on:
           - nightly
           - test
           - all
+      ref:
+        description: 'Reference to checkout, defaults to empty'
+        default: ""
+        type: string
 
 jobs:
   win:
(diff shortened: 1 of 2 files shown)


seemethere (approved):

seemethere (changes requested):
I think this might be wrong, do you actually want this? You still need some way to test this on pull requests here in builder wouldn't you?

If you hard code this here you wouldn't actually be able to test any changes

atalman:
> I think this might be wrong, do you actually want this? You still need some way to test this on pull requests here in builder wouldn't you?
> 
> If you hard code this here you wouldn't actually be able to test any changes

Hm, we normally run these on workflow_call or workflow_dispatch, let me see what I can do here

seemethere:
> > I think this might be wrong, do you actually want this? You still need some way to test this on pull requests here in builder wouldn't you?
> > If you hard code this here you wouldn't actually be able to test any changes
> 
> Hm, we normally run these on workflow_call or workflow_dispatch, let me see what I can do here

oh if that's the case then maybe we don't, but just wanted to call that out as a particular deficiency

seemethere (approved):

seemethere (approved):

malfet:
Hmm, how are we testing those workflows?
As https://github.com/pytorch/builder/actions/runs/3519117049/jobs/5898729279 indicates that our Python-1.13+CUDA-11.7  PyPI wheels are not what we claim they are? Is this something we must fix ASAP? Or just a false positive and all those results could be ignored?

PR #1215, opened 2022-12-01
Truncate RECORD instead of delete

groodt:
Fix for https://github.com/pytorch/vision/issues/6869

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -387,10 +387,10 @@ for pkg in /$WHEELHOUSE_DIR/torch*linux*.whl /$LIBTORCH_HOUSE_DIR/libtorch*.zip;
     record_file=$(echo $(basename $pkg) | sed -e 's/-cp.*$/.dist-info\/RECORD/g')
     if [[ -e $record_file ]]; then
         echo "Generating new record file $record_file"
-        rm -f $record_file
+        : > "$record_file"
         # generate records for folders in wheel
         find * -type f | while read fname; do
-            echo $(make_wheel_record $fname) >>$record_file
+            make_wheel_record "$fname" >>"$record_file"
         done
     fi
 
--- a/release/pypi/prep_binary_for_pypi.sh
+++ b/release/pypi/prep_binary_for_pypi.sh
@@ -88,10 +88,10 @@ for whl_file in "$@"; do
             record_file="${dirname_dist_info_folder}/${basename_dist_info_folder/${version_with_suffix}/${version_no_suffix}}/RECORD"
             if [[ -e $record_file ]]; then
                 echo "Generating new record file $record_file"
-                rm -f $record_file
+                : > "$record_file"
                 # generate records for folders in wheel
                 find * -type f | while read fname; do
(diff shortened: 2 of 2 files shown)


manywheel/build_common.sh
```diff
@@ -387,10 +387,10 @@ for pkg in /$WHEELHOUSE_DIR/torch*linux*.whl /$LIBTORCH_HOUSE_DIR/libtorch*.zip;
     record_file=$(echo $(basename $pkg) | sed -e 's/-cp.*$/.dist-info\/RECORD/g')
     if [[ -e $record_file ]]; then
         echo "Generating new record file $record_file"
-        rm -f $record_file
+        : > "$record_file"
```
weiwangmeta, line 390: why is ": > " better than "rm -f"?

release/pypi/prep_binary_for_pypi.sh
```diff
@@ -88,10 +88,10 @@ for whl_file in "$@"; do
             record_file="${dirname_dist_info_folder}/${basename_dist_info_folder/${version_with_suffix}/${version_no_suffix}}/RECORD"
             if [[ -e $record_file ]]; then
                 echo "Generating new record file $record_file"
-                rm -f $record_file
+                : > "$record_file"
```
weiwangmeta, line 91: Same question here, please educate. Also, is the permission change necessary from 100644 -> 100755?

weiwangmeta (commented):
Thanks for identifying and fixing the issue, just had some minor questions

manywheel/build_common.sh
groodt: The file needs to be truncated / cleared, not deleted. The bug is specifically by the deletion, because you can see the logic is conditional on the existence of that file existing: https://github.com/pytorch/builder/blob/89c8ff725fabf6d82302a4010717cfff488d1a25/manywheel/build_common.sh#L268

An alternative way to write `: > "$record_file"` could be `truncate -s 0 "$record_file"`. However, I'm not sure on how prevalent that approach is. Redirects (> and >>) are quite common, so thats why I used that style.

release/pypi/prep_binary_for_pypi.sh
groodt: > is the permission change necessary from 100644 -> 100755?

Ah, I guess not. I changed the permission to include the `+x` (execute bit) so that I could run the script: `./release/pypi/prep_binary_for_pypi.sh` If this script was somehow being launched another way, I'm happy to revert the permissions to `100644`. Let me know.

weiwangmeta (approved):
Thank you for the great fix!

seemethere (approved):

atalman (commented):
Look good! Would be nice to include some test in the description.

groodt:
@atalman I've added a verification script to the description.

What's the procedure for a new release of torchvision? Should I ask in the vision repo?

atalman:
@groodt the release of torchvision is done at the same time as pytorch, hence this fix should be included in the 1.13.1

PR #1218, opened 2022-12-05
Mitigate windows test failures

atalman:


atalman:
> I think this analysis is wrong: test faillures only exists for win-cuda builds, which manifested a while ago, see [pytorch/pytorch#89935](https://github.com/pytorch/pytorch/issues/89935)

The issue seems to be related to the following line:
```
call conda update --all -c=nvidia -y -c pytorch -c defaults -c numba/label/dev 
```

This is wrong from nightly channel should be pytorch-nightly, will amend this PR accordingly

atalman:
The install nightly step is required. Here is the failure when removed the install step:
https://github.com/pytorch/pytorch/actions/runs/3623448485/jobs/6110056506

malfet (approved):

malfet:
> The install nightly step is required. Here is the failure when removed the install step: https://github.com/pytorch/pytorch/actions/runs/3623448485/jobs/6110056506

Hmm, doesn't it mean that windows wheel smoke tests are testing previous nightly version rather than wheel that was just built?

windows/internal/smoke_test.bat
```diff
@@ -51,7 +51,7 @@ set "PATH=%CD%\Python%PYTHON_VERSION%\Scripts;%CD%\Python;%PATH%"
 pip install -q numpy protobuf "mkl>=2019"
 if errorlevel 1 exit /b 1
 
-if "%TEST_NIGHTLY_PACKAGE%" == "1" (
+if NOT "%PYTORCH_BUILD_VERSION%"=="%PYTORCH_BUILD_VERSION:dev=%" (
```
malfet, line 54: This is a very obscure check, could it be replaced with something like `TEST_WHEELS` or `NOT %TEST_CONDA%`?

atalman:
> > The install nightly step is required. Here is the failure when removed the install step: https://github.com/pytorch/pytorch/actions/runs/3623448485/jobs/6110056506
> 
> Hmm, doesn't it mean that windows wheel smoke tests are testing previous nightly version rather than wheel that was just built?

Yes. Refactored this code now to make sure we only test currently build file

PR #1224, opened 2022-12-07
conda: Add CUDA_HOME, cuda binaries to path

seemethere:
Adds CUDA_HOME and the cuda binaries to our path. Users were
experiencing issues building GPU binaries with these so this should make
it so that the proper environment variables are setup by default

### DIFF
--- a/conda/Dockerfile
+++ b/conda/Dockerfile
@@ -41,8 +41,12 @@ RUN bash ./install_conda.sh && rm install_conda.sh
 
 # Install CUDA
 FROM base as cuda
+ARG CUDA_VERSION=10.2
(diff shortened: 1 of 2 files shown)


conda/Dockerfile
```diff
@@ -41,8 +41,12 @@ RUN bash ./install_conda.sh && rm install_conda.sh
 
 # Install CUDA
 FROM base as cuda
+ARG CUDA_VERSION=10.2
```
atalman, line 44: Why is it 10.2 CUDA ? We pretty much deprecated it everywhere already, should we use something newer ? Like 11.6 ?

atalman (approved):
LGTM, only cuda question

PR #1230, opened 2022-12-13
PyPi binary validation and size check

izaitsevfb:
A semi-automated validation workflow for the PyPi binaries.

Allows to specify the list of binaries to repackage and validate as the matrix:
```yml
matrix:
  whl:
    - url: https://download.pytorch.org/whl/test/cu117_pypi_cudnn/torch-1.13.1%2Bcu117.with.pypi.cudnn-cp310-cp310-linux_x86_64.whl
      python: "3.10"  # python version to use for smoke tests
      upload_artifact: false # upload the repackaged binary as an artifact
...
```
The workflow will:
  * download the binaries,
  * run release/pypi/prep_binary_for_pypi.sh
  * run smoke tests on the repackaged binaries
  * display the size before and after repackaging as the workflow annotation
  * optionally upload the repackaged binaries as artifacts (for debug or promotion)

See [this run](https://github.com/pytorch/builder/actions/runs/3698006139).
Annotations:
image

---

Additionally, during evaluation of this workflow I discovered an issue with `prep_binary_for_pypi.sh`:

>   Apparently if the target file exists, `zip -qr9 "${new_whl_file}" .` will add to the file, instead of overwriting the whole file.
> For certain binary names, like `torch-1.13.1+cu117.with.pypi.cudnn-cp310-cp310-linux_x86_64.whl` the original and the repackaged binary names match, meaning the last step will merge them, creating two `dist-info` directories inside.
> See [this run for example](https://github.com/pytorch/builder/actions/runs/3697391375/jobs/6262319139#step:9:1924).

This issue is fixed by adding `rm -rf "${new_whl_file}"` before `zip -qr9 "${new_whl_file}" .` to ensure new file is created from scratch.

### DIFF
--- a/.github/workflows/validate-repackaged-binary-sizes.yml
+++ b/.github/workflows/validate-repackaged-binary-sizes.yml
@@ -1,65 +1,77 @@
 name: Validate manywheel binaries
 
+# This workflow validates the size of the manywheel binaries after repackaging for PyPi
+# Specify the direct URLs to the binaries (from https://download.pytorch.org/whl/test/torch/) in the matrix
+# along with the python version.
+#
+# The workflow will:
+#  * download the binaries,
+#  * run release/pypi/prep_binary_for_pypi.sh
+#  * run smoke tests on the repackaged binaries
+#  * display the size before and after repackaging
+
 on:
   pull_request:
 
 jobs:
   validate-binary-size:
-    runs-on: linux.4xlarge.nvidia.gpu
     strategy:
       fail-fast: false
       matrix:
         whl:
           - url: https://download.pytorch.org/whl/test/cu117_pypi_cudnn/torch-1.13.1%2Bcu117.with.pypi.cudnn-cp310-cp310-linux_x86_64.whl
-            python: 3.10
+            python: "3.10"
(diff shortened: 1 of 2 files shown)


release/pypi/prep_binary_for_pypi.sh
```diff
@@ -95,6 +95,7 @@ for whl_file in "$@"; do
             fi
         )
 
+        rm -rf "${new_whl_file}"
```
atalman, line 98: Why are we removing  this file before zipping ? I don't think this PR should modify the prep script

.github/workflows/validate-repackaged-binary-sizes.yml
```diff
@@ -0,0 +1,88 @@
...
+#
+# The workflow will:
+#  * download the binaries,
+#  * run release/pypi/prep_binary_for_pypi.sh
+#  * run smoke tests on the repackaged binaries
+#  * display the size before and after repackaging as the workflow annotation
+#  * optionally upload the repackaged binaries as artifacts (for debug or promotion)
+
+on:
+  pull_request:
```
atalman, line 15: Perhaps we should add on request, when running manually.
And add a pytorch version and cuda version as a parameter. Since we do plan on introducing cu118 soon.
Also add nightly/test as a parameter as well

atalman (approved):
LGTM, please address comments

PR #1232, opened 2022-12-14
Refactor PyTorch wheel and libtorch build scripts for ROCm

jithunnair-amd:
**MOTIVATION:**

To refactor the wheel/libtorch build scripts for ROCm to make them less verbose, more robust to library version changes from one ROCm version to the next, and less repetitive.

**IMPLEMENTATION:**

Thanks to @jataylo for the heavy lifting!

* Update to so patching for ROCm

Wildcard used in grep to grab the actual numbered so file referenced in patchelf. This allows the removal of specifying the so number in DEPS_LIST & DEPS_SONAME

This commit also adds the functionality for trimming so names to build_libtorch.sh from build_common.sh

* Refactor to remove switch statement in build_rocm.sh

This commit refactors build_rocm.sh and brings in a few major updates:
 - No longer required to specify the full .so name (with number) for ROCm libraries - The .so versions are copied and the patching code will fix the links to point to this version
 - No longer required to specify paths for ROCm libraries allowing the removal of the large switch - Paths are acquired programmatically with find
 - No longer required to specify both the path and filename for the OS specific libraries - Programmatically extract file name from the path
 - Automatically extract Tensile/Kernels files for the architectures specified in PYTORCH_ROCM_ARCH and any non-arch specific files e.g. TensileLibrary.dat

Tested in https://github.com/pytorch/pytorch/pull/90879

### DIFF
--- a/manywheel/build_common.sh
+++ b/manywheel/build_common.sh
@@ -279,9 +279,9 @@ replace_needed_sofiles() {
     find $1 -name '*.so*' | while read sofile; do
         origname=$2
         patchedname=$3
-        if [[ "$origname" != "$patchedname" ]]; then
+        if [[ "$origname" != "$patchedname" ]] || [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
             set +e
-            $PATCHELF_BIN --print-needed $sofile | grep $origname 2>&1 >/dev/null
+            origname=$($PATCHELF_BIN --print-needed $sofile | grep "$origname*") 
             ERRCODE=$?
             set -e
             if [ "$ERRCODE" -eq "0" ]; then
--- a/manywheel/build_libtorch.sh
+++ b/manywheel/build_libtorch.sh
@@ -228,6 +228,11 @@ fname_with_sha256() {
     fi
 }
 
+fname_without_so_number() {
+    LINKNAME=$(echo $1 | sed -e 's/\.so.*/.so/g')
+    echo "$LINKNAME"
+}
+
 make_wheel_record() {
     FPATH=$1
     if echo $FPATH | grep RECORD >/dev/null 2>&1; then
@@ -277,8 +282,12 @@ for pkg in /$LIBTORCH_HOUSE_DIR/libtorch*.zip; do
             if [[ "$filepath" != "$destpath" ]]; then
                 cp $filepath $destpath
             fi
-
-            patchedpath=$(fname_with_sha256 $destpath)
+            
+            if [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
+                patchedpath=$(fname_without_so_number $destpath)
+            else
+                patchedpath=$(fname_with_sha256 $destpath)
+            fi
             patchedname=$(basename $patchedpath)
             if [[ "$destpath" != "$patchedpath" ]]; then
                 mv $destpath $patchedpath
@@ -292,9 +301,9 @@ for pkg in /$LIBTORCH_HOUSE_DIR/libtorch*.zip; do
             find $PREFIX -name '*.so*' | while read sofile; do
                 origname=${DEPS_SONAME[i]}
                 patchedname=${patched[i]}
-                if [[ "$origname" != "$patchedname" ]]; then
+                if [[ "$origname" != "$patchedname" ]] || [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
                     set +e
-                    $PATCHELF_BIN --print-needed $sofile | grep $origname 2>&1 >/dev/null
+                    origname=$($PATCHELF_BIN --print-needed $sofile | grep "$origname*") 
                     ERRCODE=$?
                     set -e
                     if [ "$ERRCODE" -eq "0" ]; then
--- a/manywheel/build_rocm.sh
+++ b/manywheel/build_rocm.sh
@@ -2,7 +2,8 @@
 
 set -ex
 
-export MAGMA_HOME=/opt/rocm/magma
+export ROCM_HOME=/opt/rocm
+export MAGMA_HOME=$ROCM_HOME/magma
 # TODO: libtorch_cpu.so is broken when building with Debug info
 export BUILD_DEBUG_INFO=0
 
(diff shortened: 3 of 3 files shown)


jithunnair-amd:
@atalman @malfet Please review this PR; we'll be following this up with ROCm5.4 PRs.

manywheel/build_common.sh
```diff
@@ -279,9 +279,9 @@ replace_needed_sofiles() {
     find $1 -name '*.so*' | while read sofile; do
         origname=$2
         patchedname=$3
-        if [[ "$origname" != "$patchedname" ]]; then
+        if [[ "$origname" != "$patchedname" ]] || [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
             set +e
-            $PATCHELF_BIN --print-needed $sofile | grep $origname 2>&1 >/dev/null
+            origname=$($PATCHELF_BIN --print-needed $sofile | grep "$origname*")
```
malfet, line 284: Can you please explain why this change is needed? Looks like previously it was designed to just print errors, but now it would still print them but also preserve all the dependencies as origname?

malfet (commented):
@jithunnair-amd can you please create a PR against pytorch to test the change

manywheel/build_common.sh
jithunnair-amd: @jataylo Would you mind explaining the reasoning behind this change, maybe with an example?

jataylo: Hi @malfet @jithunnair-amd long message but I thought some more context would help.

This change is required due to the new way we specify the so files in `build_rocm.sh` and specifically the change to `DEPS_SONAME`.

**build_rocm.sh**
Previously this was a hard-coded list of full so library names:
```
DEPS_SONAME=(
    “libMIOpen.so.1”
    “libamdhip64.so.5”
     …
)
```
This PR makes a change to no longer require the full numbered so library name be specified. For example **libMIOpen.so** instead of l**ibMIOpen.so.1** allowing us to greatly simplify the code instead of manually listing dependencies for each ROCm version.
```
DEPS_SONAME=(
    ${ROCM_SO_FILES[*]}
    ${OS_SO_FILES[*]}
)
ROCM_SO_FILES=(
    "libMIOpen.so"
    "libamdhip64.so"
    ...
)
```

**build_common.sh**
In `build_common.sh` another change is required to support this feature, stepping back to explain this context. In build_common.sh the library name is patched, which in our case to remove the so number from a specified so library name (this is still needed as some libraries are still using the full numbered so library name)
```
# ROCm workaround for roctracer dlopens
if [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
    patchedpath=$(fname_without_so_number $destpath)
else
    patchedpath=$(fname_with_sha256 $destpath)
fi
```
This appends all the patched library names (trimmed so numbers) to the patched array.

The actual patching code is contained in ` replace_needed_sofiles`  which takes in the new DEPS_SONAME and the new patched library names and patches the dependencies.
```
echo "patching to fix the so names to the hashed names"
for ((i=0;i<${#DEPS_LIST[@]};++i)); do
    replace_needed_sofiles $PREFIX ${DEPS_SONAME[i]} ${patched[i]}
```

Originally inside ` replace_needed_sofiles()`  each so file in torch is processed and if `$origname` & `$patchedname` do not match, then the dependencies of the library are printed. If $origname exists in the dependencies the torch library dependencies are patched to point to the none so-numbered version

```
replace_needed_sofiles() {
    find $1 -name '*.so*' | while read sofile; do
        origname=$2
        patchedname=$3
        if [[ "$origname" != "$patchedname" ]]; then
            set +e
            $PATCHELF_BIN --print-needed $sofile | grep $origname 2>&1 >/dev/null
            ERRCODE=$?
            set -e
            if [ "$ERRCODE" -eq "0" ]; then
                echo "patching $sofile entry $origname to $patchedname"
                $PATCHELF_BIN --replace-needed $origname $patchedname $sofile
            fi
        fi
    done
}
```

With this PR we need a change because we are no longer specifying the full numbered so name for all libraries in DEPS_SONAME, but the links to torch dependencies will still be to the numbered versions. To get around this on ROCm we will bypass this condition, but retain current functionality for other platforms.
```
if [[ "$origname" != "$patchedname" ]] || [[ "$DESIRED_CUDA" == *"rocm"* ]]; then
```

The next change is to grep for the none numbered so file in the dependencies, extracting the numbered so file from patchelf –print-needed by grepping for $origname* instead of $origname. If $origname is found in –print-needed then ERRCODE=0 and the patching occurs otherwise no patching occurs.
```
origname=$($PATCHELF_BIN --print-needed $sofile | grep "$origname*")
```

**An example**
(prior to change)
```
origname=libMIOpen.so.1
patchedname=libMIOpen.so 
/usr/local/lib/python3.8/dist-packages/torch/lib# patchelf --print-needed libtorch_hip.so
…
libtorch_hip.so 
libc10_hip.so 
libamdhip64.so.5 
libMIOpen.so.1 
librccl.so.1 
libroctx64.so 
librocblas.so.0 
libhipfft.so 
libhiprand.so.1 
libhipsparse.so.0 
libmagma.so 
libc10.so 
libtorch_cpu.so 
libpthread.so.0 
libm.so.6 
librt.so.1 
libstdc++.so.6 
libgomp.so 
libgcc_s.so.1 
libc.so.6 
ld-linux-x86-64.so.2

ERRCODE=0 (as libMIOpen.so.1 is found in libtorch_hip.so)
patchelf -replace-needed libMIOpen.so.1 libMIOpen.so libtorch_hip.so
```
(Post PR)
```
origname=libMIOpen.so
patchedname=libMIOpen.so 
/usr/local/lib/python3.8/dist-packages/torch/lib# patchelf --print-needed libtorch_hip.so
…
libtorch_hip.so 
libc10_hip.so 
libamdhip64.so.5 
libMIOpen.so.1 
librccl.so.1 
libroctx64.so 
librocblas.so.0 
libhipfft.so 
libhiprand.so.1 
libhipsparse.so.0 
libmagma.so 
libc10.so 
libtorch_cpu.so 
libpthread.so.0 
libm.so.6 
librt.so.1 
libstdc++.so.6 
libgomp.so 
libgcc_s.so.1 
libc.so.6 
ld-linux-x86-64.so.2

origname=$(patchelf --print-needed libtorch_hip.so | grep “$origname*”)
echo $origname 
libMIOpen.so.1

ERRCODE=0 (as libMIOpen.so.1 is found in libtorch_hip.so)
patchelf -replace-needed libMIOpen.so.1 libMIOpen.so libtorch_hip.so
```

In short this does not change the current functionality as the addition of the wildcard in the grep will not have any effect if the full so name is already given but will help in our case as we can programmatically get the full so numbered version in the dependencies and patch accordingly.

jithunnair-amd: @malfet I think the above comment clarifies that the `origname=$(patchelf --print-needed libtorch_hip.so | grep “$origname*”)` should have no impact on CUDA. We went with this implementation since it looks/reads cleaner. If you prefer conditionalizing this code for CUDA to retain the old code as well, we can do that, but it will look a bit less clean.

atalman: @jithunnair-amd We need to test this change before committing. Can you please commit this change to  your fork main branch and then open PR in pytorch/pytorch and change pytorch/builder reference to your forked main branch.

jithunnair-amd: @atalman @malfet I think the CI on that PR looks good, so can we consider this properly tested?

atalman (approved):
LGTM

malfet (approved):
Thank you for the explanation, looks good to me

PR #1238, opened 2022-12-14
Adding more OS's to validate domain library workflow

atalman:
Adding more OS's to validate domain library workflow

### DIFF
--- a/.github/workflows/validate-domain-library.yml
+++ b/.github/workflows/validate-domain-library.yml
@@ -36,43 +36,27 @@ on:
         type: string
 
 jobs:
-  generate-linux-conda-matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-domain-library.yml
```diff
@@ -50,7 +50,23 @@ jobs:
     with:
       package-type: wheel
       os: linux
-      channel: ${{ inputs.channel || 'nightly' }}
+      channel: ${{ inputs.channel }}
+      with-cuda: disable
```
weiwangmeta, line 54: is disable with-cuda intentional here?

atalman: Yes. For now. Will add this as input parameter later

weiwangmeta (approved):
LGTM

PR #1240, opened 2022-12-15
Build Windows binaries with Visual Studio 2022 Build Tools

Blackhex:
This builds Windows binaries with VS 2022 and this PR is a preparation for adding Windows ARM64 support. The changes are enabled and tested by this PR pytorch/pytorch#90855.

cc @jeanschmidt

Together with  pytorch/pytorch#90855 closes pytorch/pytorch#87695.


jeanschmidt (approved):

windows/internal/vc_install_helper.bat
```diff
@@ -1,7 +1,17 @@
+if "%VC_YEAR%" == "2017" powershell windows/internal/vs2017_install.ps1
```
atalman, line 1: I don't think VS2017 is used anywhere now. We should probably keep 2 versions 2022 and 2019 (if we will need to fallback/revert) .  In this PR we should not add additional cases for VS2017. Please remove them. I think follow up PR to remove all the VS2017 artifacts will be a good idea.

atalman (changes requested):
Looks good and test PR looks good as well. Please address comment about vs2017

conda/vs2022/activate.bat
```diff
@@ -0,0 +1,51 @@
...
+SET MSSdk=1
+
+SET "VS_VERSION=17.4"
+SET "VS_MAJOR=17"
+SET "VC_YEAR=2022"
+SET "VC_VERSION_LOWER=17"
+set "VC_VERSION_UPPER=18"
+
+set "MSYS2_ARG_CONV_EXCL=/AI;/AL;/OUT;/out"
+set "MSYS2_ENV_CONV_EXCL=CL"
```
malfet, line 12: Why lowercase set here? Do we have some sort of linter
```suggestion
SET "VC_VERSION_UPPER=18"

SET "MSYS2_ARG_CONV_EXCL=/AI;/AL;/OUT;/out"
SET "MSYS2_ENV_CONV_EXCL=CL"
```

malfet (changes requested):
Please do not add a new `.bat` files, as there are really no error handling in batch. I.e. all the new files in `conda/vs2022` folder should be either Python or PowerShell ones

But otherwise, looks good to me.

conda/vs2022/activate.bat
Blackhex: I've basically copied what's in the vs2017 folder already. I can certainly unify the casing. So the upper-case is the code style adopted? Or rewrite to something else than Batch.

malfet: My point is more that we should not keep copying those files over and over, but finally switch to PowerShell while doing the upgrade.

windows/internal/vc_install_helper.bat
Blackhex: I've removed the VS2017 installation from the build scripts.

conda/vs2022/activate.bat
Blackhex: Unless I am missing something, those Batch files are needed for the `conda-build` recipe and it does not seem to support PowerShell yet (conda/conda-build#3967). I've at least removed the recipe for VS2017 and unified the code style casing.

Blackhex:
Ready for the next round of the review.

Blackhex:
@malfet Any update on this? Would moving all the code from batch scripts to their corresponding PowerShell scripts and excecuting those PowerShell scripts from the original batch scritpts be a viable solution for you? E.g. `activate.bat` would just call `activate.ps1` doing the thing.

Blackhex:
Or do we need the conda package activation script (and the script that is copying it during installation) at all considering what is written at https://docs.conda.io/projects/conda-build/en/latest/resources/activate-scripts.html:

> It is generally recommended to avoid using activate scripts when another option is possible because people do not always activate environments the expected way and these packages may then misbehave.

and the fact that this activation script just activates VS environment? Is it a good practice to activate VS environment when conda environment is activated with PyTorch package installed? I'd say it is not.

atalman (approved):

malfet (approved):

PR #1255, opened 2022-12-29
Remove pylief dependency

atalman:
Remove pylief dependency. 

This is to unblock macos conda builds that are failing: https://github.com/pytorch/pytorch/actions/runs/3799171656/jobs/6461449086

Test PR:
https://github.com/pytorch/pytorch/pull/91507
https://github.com/pytorch/pytorch/actions/runs/3801621293/jobs/6466278871

I don't observe any difference in building times vs previously passing build 1hr:07-1hr:15 min:
https://github.com/pytorch/pytorch/actions/runs/3756015842/jobs/6381619694

Looks like the issue caused by long building times is resolved now as per Test PR:
https://github.com/pytorch/pytorch/issues/58534

### DIFF
--- a/conda/build_pytorch.sh
+++ b/conda/build_pytorch.sh
@@ -207,8 +207,6 @@ if [[ "$(uname)" == 'Darwin' ]]; then
         rm "$miniconda_sh"
     export PATH="$tmp_conda/bin:$PATH"
     retry conda install -yq conda-build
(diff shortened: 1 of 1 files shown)


malfet (approved):
LGTM, though can we please pin py-lief in another file?

atalman:
> LGTM, though can we please pin py-lief in another file?

HI @malfet sure I can pin py-lief. But do we need it at all ? Looks like there is no regression in terms of building time any more.

PR #1262, opened 2023-01-04
Adding py311 validations for wheel builds

atalman:
Adding python 3.11 validations for nightly testing.

Please note: Compiling python```--with-openssl``` is needed in CentOS, this is what we are running in Docker.

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -11,14 +11,28 @@ if [[ ${PACKAGE_TYPE} == "libtorch" ]]; then
     curl ${INSTALLATION} -o libtorch.zip
     unzip libtorch.zip
 else
-    conda create -y -n ${ENV_NAME} python=${DESIRED_PYTHON} numpy pillow
-    conda activate ${ENV_NAME}
-    export CONDA_LIBRARY_PATH="$(dirname $(which python))/../lib"
-    export LD_LIBRARY_PATH=$CONDA_LIBRARY_PATH:$LD_LIBRARY_PATH
-    INSTALLATION=${INSTALLATION/"conda install"/"conda install -y"}
-    eval $INSTALLATION
-    python  ./test/smoke_test/smoke_test.py
-    if [[ ${TARGET_OS} != 'macos' && ${TARGET_OS} != 'windows' ]]; then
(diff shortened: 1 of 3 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -11,14 +11,28 @@ if [[ ${PACKAGE_TYPE} == "libtorch" ]]; then
...
+        sudo yum -y install openssl-devel libssl-dev bzip2-devel libffi-devel
+        sudo yum -y groupinstall "Development Tools"
+        export PYTHON_PATH="/opt/_internal/cpython-3.11.0/bin"
+        export PIP_PATH="${PYTHON_PATH}/pip"
+        export PIP_INSTALLATION="${INSTALLATION/pip3/"$PIP_PATH"}"
+        export WITH_OPENSSL="/opt/openssl"
+        ./common/install_cpython.sh
+        eval ${PYTHON_PATH}/python --version
+        eval ${PIP_INSTALLATION}
+        eval ${PYTHON_PATH}/python ./test/smoke_test/smoke_test.py --package torchonly
```
albanD, line 25: The plan is to do all packages and the full check_binary in a follow up right?

common/install_cpython.sh
```diff
@@ -40,7 +40,11 @@ function do_cpython_build {
     mkdir -p ${prefix}/lib
 
     # -Wformat added for https://bugs.python.org/issue17547 on Python 2.6
-    CFLAGS="-Wformat" ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
+    if [[ -z  ${WITH_OPENSSL} ]]; then
+        CFLAGS="-Wformat" ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
```
albanD, line 44: btw if you have issues with pip missing for some versions, you can add `--with-ensurepip=install` to this

albanD (approved):
SGTM !

.github/scripts/validate_binaries.sh
atalman: Yes, once we have torchvision and torchaudio with 3.11 we will remove torchonly option

malfet:
This broke manywheel generation a while back, and nobody bothered to check 👎

PR #1266, opened 2023-01-11
Fix validation domain library workflow

atalman:
Pass matrix to test-infra jobs so that MATRIX_* variables are initialzied


weiwangmeta (commented):
LGTM!

PR #1289, opened 2023-01-31
Add pypi with cudnn package test

atalman:
Add pypi with cudnn package test. This is the package that will be uploaded to pypi.


weiwangmeta (approved):
LGTM

PR #1225, opened 2022-12-09
Upgrade nightly wheels to rocm5.4.2

BLOrange-AMD:

PR #1296, opened 2023-02-08
rocm libtorch prebuild magma; fix manylinux cmake version

jeffdaily:
Speed up rocm libtorch image generation by providing a base rocm image with magma already installed.

#1225 inadvertently reverted the cmake version from 3.18 to 3.17 and broke wheel builds.  This is fixed.


atalman (approved):
LGTM

PR #1297, opened 2023-02-08
Adding limit windows builds logic

atalman:
Adding limit windows builds logic
Cleaning up conda environment after validation has been run


.github/scripts/validate_binaries.sh
```diff
@@ -41,5 +41,8 @@ else
         fi
 
         python  ./test/smoke_test/smoke_test.py
+
+        conda deactivate
```
weiwangmeta, line 45: Should these be aligned with the else, i.e. directly above the fi?

weiwangmeta (commented):
Please clarify the errors if the conda deactivate is not used and why those should be put in the else statement.

seemethere (approved):
LGTM, let's address @weiwangmeta's comments prior to merge

.github/scripts/validate_binaries.sh
weiwangmeta: Oh, nvm. It was only created inside this else statement. So it should be correct.

weiwangmeta (approved):
Please wait for green signal and merge.

PR #1304, opened 2023-02-10
Python 3.11 validation workflow tests

atalman:
This is general Python 3.11 validation worklfow test of core only for now.


weiwangmeta (approved):
LGTM

PR #1327, opened 2023-03-01
Add filelock and networkx deps

malfet:
To match dependencies for wheel files defined in https://github.com/pytorch/pytorch/blob/ed1957dc1989417cb978d3070a4e3d20520674b4/setup.py#L1021-L1024

Tested in https://github.com/pytorch/pytorch/pull/95780


weiwangmeta (approved):
Looks good to me. 
These seem to be a superset of https://github.com/pytorch/pytorch/blob/release/2.0/setup.py#L1022-L1026

PR #1331, opened 2023-03-06
Upgrade cmake version to 3.22.1 to build triton

huydhn:
I try to upgrade cmake version in builder images to 3.22.1, same as the version used by PyTorch CI, to build triton.  The newer triton pinned commit is now complaining about cmake version https://github.com/pytorch/pytorch/actions/runs/4332758131:

* manywheel uses cmake-3.18.4 from pip
* conda uses system cmake-3.17.5

Both are upgraded to cmake-3.22.1.

I also need to pin patcheft version to the current latest ([0.17.2](https://github.com/NixOS/patchelf/releases/tag/0.17.2)) instead of cloning the latest commit from GitHub as the build starts to fail https://github.com/pytorch/builder/actions/runs/4339776911/jobs/7577682033.  We might need to investigate this further if it's important to use the latest commit for some reasons.

### Testing
Let me know if I need to do some further testing.

### DIFF
--- a/common/install_conda.sh
+++ b/common/install_conda.sh
@@ -9,5 +9,7 @@ chmod +x  Miniconda3-latest-Linux-x86_64.sh
 bash ./Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda
 rm Miniconda3-latest-Linux-x86_64.sh
 export PATH=/opt/conda/bin:$PATH
-conda install -y conda-build anaconda-client git ninja
+# cmake-3.22.1 from conda, same as the one used by PyTorch CI. The system cmake
+# is too old to build triton
+conda install -y conda-build anaconda-client git ninja cmake=3.22.1
 conda remove -y --force patchelf
--- a/common/install_patchelf.sh
+++ b/common/install_patchelf.sh
@@ -2,8 +2,8 @@
 
 set -ex
 
-# Pin the version to latest release 0.17.2, newer version starts failing
-# to build on the current image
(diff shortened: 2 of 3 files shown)


atalman (approved):
LGTM: Would be good to make sure it does not break master or tomorrow nightly builds.
Current nightly state is green: https://hud.pytorch.org/hud/pytorch/pytorch/nightly
After merging, can you please manually rerun nightly builds ? And see if they are still green ?

huydhn:
Raise a request issue to GitHub https://support.github.com/ticket/personal/0/2043964 (Weird, why it shows personal instead of enterprise, may be because I don't have enterprise GitHub account, hmm)

malfet:
My problem with this change, is that devs can sneak newer cmake features in CMakeLists.txt without updating minimum supported cmake version.

huydhn:
> My problem with this change, is that devs can sneak newer cmake features in CMakeLists.txt without updating minimum supported cmake version.

Note my chat with @malfet here for future reference:

* Essentially we need to validate that builds are succeeding with 3.18.5, which is the current min version for PyTorch
* Triton enforces 3.20 since this commit https://github.com/openai/triton/commit/77c145cec8b6692193f692aed48d22d2eea70fd0

```
# Introduced in cmake 3.20
# https://cmake.org/cmake/help/latest/policy/CMP0116.html
cmake_policy(SET CMP0116 OLD)
```

* The above change is better to be guarded against CMAKE version (thank @malfet for pointing this out to triton folks).  
* Once they have fix it and allow us to use older cmake version, I will revert this commit.

PR #1334, opened 2023-03-06
Switch deprecated ubuntu-18.04 runner to ubuntu-latest

huydhn:
GitHub ubuntu-18.04 runner will be deprecated soon https://github.com/actions/runner-images/issues/6002, so switching to our ~~self-hosted 2xlarge~~ ephemeral ubuntu-latest runner.

I will rerun all nightly after the Docker images are rebuild.

### DIFF
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build-docker:
-    runs-on: linux.c5.4xlarge.ephemeral
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         cuda_version: ["11.6", "11.7", "11.8", "cpu"]
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build-docker:
-    runs-on: linux.2xlarge
+    runs-on: linux.c5.4xlarge.ephemeral
     strategy:
       matrix:
         cuda_version: ["11.6", "11.7", "11.8", "cpu"]
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build-docker:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         cuda_version: ["11.6", "11.7", "11.8", "cpu"]
--- a/.github/workflows/build-libtorch-images.yml
+++ b/.github/workflows/build-libtorch-images.yml
@@ -25,7 +25,7 @@ env:
 
 jobs:
   build-docker-cuda:
-    runs-on: linux.2xlarge
+    runs-on:  linux.large.ephemeral
     strategy:
       matrix:
         cuda_version: ["11.8", "11.7", "11.6"]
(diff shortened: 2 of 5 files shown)


.github/workflows/build-libtorch-images.yml
```diff
@@ -25,7 +25,7 @@ env:
 
 jobs:
   build-docker-cuda:
-    runs-on: ubuntu-18.04
+    runs-on: linux.2xlarge
```
malfet, line 28:
```suggestion
    runs-on:  linux.large.ephemeral
```

.github/workflows/build-manywheel-images.yml
```diff
@@ -27,7 +27,7 @@ env:
 
 jobs:
   build-docker-cuda:
-    runs-on: ubuntu-18.04
+    runs-on: linux.2xlarge
```
malfet, line 30:
```suggestion
    runs-on: linux.c5.4xlarge.ephemeral
```

huydhn:
> If I understand correctly, for security purposes, all binary builds must be executed on ephemeral runners (which GitHub stock runners guarantee, but for self-hosted runners we need to use `.ephemeral` types)

AFAIK, this is only to build the Docker images, which are then used to build our binaries.  So this is not directly involved with the latter build process.  I choose `2xlarge` here mainly because the existing conda build workflow already uses it https://github.com/pytorch/builder/blob/main/.github/workflows/build-conda-images.yml#L26

But then, I agree that it's better to be safe than sorry here, especially if we have ephemeral runners to use.  So let me switch them all to ephemeral runners

huydhn:
Hmm, the ephemeral runners queue is going to take forever https://github.com/pytorch/builder/actions/runs/4350250859, is this a possibility to consider non-ephemeral runners here instead? I think I will go with `ubuntu-latest` in the meantime as it's the safe and working choice

.github/workflows/build-libtorch-images.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: |
           libtorch/build_docker.sh
   build-docker-rocm:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
malfet, line 47:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/build-libtorch-images.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         run: |
           libtorch/build_docker.sh
   build-docker-cpu:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
malfet, line 66:
```suggestion
    runs-on: ubuntu-22.04
```

malfet (approved):
Please use `ubuntu-22.04` instead of latest. And keep `linux.2xlarge` in place (you can add `[self-hosted, linux.2xlarge]` to be more explicit that this is not run on GH infrastructure)

.github/workflows/build-conda-images.yml
```diff
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build-docker:
-    runs-on: linux.2xlarge
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
huydhn: This was probably an oversight in the past.  I couldn't figure out why we use non-ephemeral `linux.2xlarge` for conda build while `ubuntu-VERSION` for wheel.  It makes sense to have both running in the same type of non-ephemeral runner?

PR #1362, opened 2023-03-17
[aarch64] update readme with the "--enable-mkldnn" option

snadampal:
This needs to be enabled for official wheel building.


malfet (approved):

malfet:
I think better approach would be to flip the switch, but let me do it in the followup PR

snadampal:
Thanks @malfet . I see we need to do some more cleanup
1. flipping the mkldnn flag
2. making docker option the default (anyway host builds are not supported)
3. make ubuntu 20.04 or 22.04 the default, 18.04 might reach EOL soon.

PR #1368, opened 2023-03-28
add magma build for CUDA12.1

ptrblck:
This PR adds magma builds for CUDA 12.1 and removes CUDA 11.6 builds.
Additionally, the `CMake.src` file needs to be patched to drop the deprecated `fermi` architecture.
Note that `fermi` is sm_20/sm_21 and is already not supported, so no changes should be needed for CUDA11.x builds.

CC @atalman @malfet

### DIFF
--- a/.github/workflows/build-magma-linux.yml
+++ b/.github/workflows/build-magma-linux.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: linux.2xlarge
     strategy:
       matrix:
-        cuda_version: ["118", "117", "116"]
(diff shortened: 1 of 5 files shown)


ptrblck:
This needs to land after the docker images are available from https://github.com/pytorch/builder/pull/1367

atalman:
If this workflow is successful we should retry running the checks on this PR: https://github.com/pytorch/builder/actions/runs/4546265658

atalman (approved):
LGTM

PR #1376, opened 2023-04-03
Windows CUDA 12.1 changes

ptrblck:
CC @atalman @malfet


atalman (approved):
LGTM
We should perhaps think about followup where to migrate https://github.com/ptrblck/magma_win.git project.

PR #1398, opened 2023-05-03
Do not cleanup MSVC and CUDA on Windows non-ephemeral runners

huydhn:
This step cleans up MSVC and CUDA every time after a binary build job, thus breaking the next job on the new Windows non-ephemeral runner.

Also fix the wrong `circleci` username as the job is now running on GitHub


kit1980 (approved):

malfet (approved):
Wow!

atalman (approved):

PR #1412, opened 2023-05-30
Fix PyTorch init time due to inefficient libdrm search for amdgpu.ids

### LINKED ISSUE #1410: Pytorch ROCm Init is slow (3+ minutes)
After investigation of  https://github.com/ROCmSoftwarePlatform/pytorch/issues/1232
we found the root cause of the issue to be https://github.com/pytorch/builder/blob/main/common/install_rocm_drm.sh#L100

* The code looks for "amdgpu.ids" at the root of the executable (removing /bin/exec) (i.e /path/to/bin/python => /path/to)

1. The `check_for_location_of_amdgpuids` always returns 0 so the search is never cancelled once the first file is found.
2. because the search is never cancelled the entire python installation is walked through 
3. the file is not even installed close to the python installation so looking there is pointless anyway
4. the amgpu.ids is just a mapping of product id to marketing names; the features is far from critical but takes a significant amount of time for no reasons.
5. other part of the code will directly load `/opt/amdgpu/share/libdrm/amdgpu.ids ` so why bother looking for it if we know where it is.

jeffdaily:
Fixes #1410. 
The libdrm replacement for ROCm images will first search the typical location for the amdgpu.ids file. If failed, it will search as it did before in the python install location.

### DIFF
--- a/common/install_rocm_drm.sh
+++ b/common/install_rocm_drm.sh
@@ -32,13 +32,13 @@ pushd drm
 ###########################
 patch -p1 <<'EOF'
 diff --git a/amdgpu/amdgpu_asic_id.c b/amdgpu/amdgpu_asic_id.c
-index a5007ffc..a3627529 100644
+index a5007ffc..13fa07fc 100644
 --- a/amdgpu/amdgpu_asic_id.c
 +++ b/amdgpu/amdgpu_asic_id.c
 @@ -22,6 +22,13 @@
   *
   */
- 
+
 +#define _XOPEN_SOURCE 700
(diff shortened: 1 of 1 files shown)


jithunnair-amd (approved):

jeffdaily:
CUDA image failed to build.  Unrelated to these changes.

jithunnair-amd:
@atalman Please merge this PR to update the ROCm docker images

malfet (approved):

malfet:
Oops, sorry about the botched commit description

PR #1414, opened 2023-06-02
formalize the ROCm image pre-build process

jeffdaily:
ROCm must build magma and miopen from source. This takes considerable time and also exceeds allowed disk space on the GitHub Action runner.

ROCm images are now a two-stage process. For example, first run:

GPU_ARCH_VERSION=5.5 ./manywheel/build_docker_rocm_prebuild.sh

This builds and pushes the image:

rocm/dev-centos-7:5.5-magma-miopen-staging

Then the GHA workflow can proceed as usual.

### DIFF
--- a/manywheel/Dockerfile
+++ b/manywheel/Dockerfile
@@ -153,19 +153,40 @@ RUN rm -rf /usr/local/cuda-${BASE_CUDA_VERSION}
 COPY --from=cuda     /usr/local/cuda-${BASE_CUDA_VERSION}  /usr/local/cuda-${BASE_CUDA_VERSION}
 COPY --from=magma    /usr/local/cuda-${BASE_CUDA_VERSION}  /usr/local/cuda-${BASE_CUDA_VERSION}
 
-FROM cpu_final as rocm_final
+### Special target for ROCm for building magma and miopen from source ahead of time.
+### For this target, GPU_IMAGE is rocm/centos-7:${ROCM_VERSION}-complete
+FROM ${GPU_IMAGE} as rocm_prebuild
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
 ARG ROCM_VERSION=3.7
 ARG PYTORCH_ROCM_ARCH
 ENV PYTORCH_ROCM_ARCH ${PYTORCH_ROCM_ARCH}
-# Install ROCm
-ADD ./common/install_rocm.sh install_rocm.sh
-RUN ROCM_VERSION=${ROCM_VERSION} bash ./install_rocm.sh && rm install_rocm.sh
+# Install ROCm --- not needed, GPU_IMAGE contains full rocm install
+#ADD ./common/install_rocm.sh install_rocm.sh
+#RUN ROCM_VERSION=${ROCM_VERSION} bash ./install_rocm.sh && rm install_rocm.sh
+# Install patched libdrm
 ADD ./common/install_rocm_drm.sh install_rocm_drm.sh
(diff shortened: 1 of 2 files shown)


malfet (changes requested):
Please add CI to do that. Also, there are no harm in picking a bigger runner for the job (as Docker images are rebuilt very infrequently)
Also, just curios, why MAGMA built mechanism for ROCm and CUDA are diverged?

jithunnair-amd:
> Please add CI to do that. Also, there are no harm in picking a bigger runner for the job (as Docker images are rebuilt very infrequently) Also, just curios, why MAGMA built mechanism for ROCm and CUDA are diverged?

@malfet Do you mean "add CI to build rocm/dev-centos-7:5.5-magma-miopen-staging" image? That docker image wouldn't be in a PyTorch docker repository.
As for Magma, the CUDA flow installs magma from a conda package, while the ROCm flow builds magma from source.

manywheel/Dockerfile
```diff
@@ -142,7 +142,7 @@ RUN rpm -e cmake || true
 # cmake-3.18.4 from pip
 RUN yum install -y python3-pip && \
     python3 -mpip install cmake==3.18.4 && \
-    ln -s /usr/local/bin/cmake /usr/bin/cmake
+    ln -sf /usr/local/bin/cmake /usr/bin/cmake
```
jithunnair-amd, line 145: @jeffdaily Can you please elaborate why the force link is needed as part of this change?

jithunnair-amd:
> > @malfet Do you mean "add CI to build rocm/dev-centos-7:5.5-magma-miopen-staging" image? That docker image wouldn't be in a PyTorch docker repository.
> 
> @jithunnair-amd and that's a problem, isn't it? Because if somebody needs to fix/update magma, they'll have to ask you guys to update the basic container.
> 

I will say that we actually considered using one of our ROCm GHA runners for the builder CI jobs, but picking a bigger CPU-only GHA runner seems like a fine idea too. @jeffdaily, should we try the bigger GHA runner to see if we can avoid this two-stage docker build process?

> > As for Magma, the CUDA flow installs magma from a conda package, while the ROCm flow builds magma from source.
> 
> Yes, I know the mechanics, but the question is: why? I.e. magma for CUDA is build by the following [workflow](https://github.com/pytorch/builder/blob/main/.github/workflows/build-magma-linux.yml), why can't rocm build of magma be added to it?

Actually, we have been working on this recently, trying to take a look at the CUDA magma build flow to see if we can apply it to ROCm magma as well. It's a WIP.

malfet:
@jithunnair-amd I'm not saying that ROCm magma builds need to follow CUDA magma builds or vice versa, but it would be nice if workflows can be kept as close to each other as possible.

manywheel/Dockerfile
jeffdaily: The pre-build of the magma-miopen-staging image requires cmake 3 and various dockerfile targets all try to pip install cmake and create the same symlink.  When we layer these different targets the symlink already exists and will error unless forced.  The pip install will exit cleanly as already complete.

jithunnair-amd:
@malfet What are the runner instance names to use for larger diskspace/CPU? This doesn't indicate specific labels: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners?

malfet:
@jithunnair-amd we have self-hosted runners running on AWS which are defined in https://github.com/pytorch/test-infra/blob/main/.github/scale-config.yml

jeffdaily:
Closing in favor of https://github.com/pytorch/builder/pull/1418.

PR #1418, opened 2023-06-07
One-step ROCm manywheel/libtorch docker build

jithunnair-amd:
Salient points:
* Using linux.12x.large for beefier CPU and more disk space
* Using `-complete` docker images as base (which contain full ROCm installation) and removing ROCm install step as a result
* Reinstating MIOpen and magma build from source

cc @jeffdaily

### DIFF
--- a/.github/workflows/build-libtorch-images.yml
+++ b/.github/workflows/build-libtorch-images.yml
@@ -44,7 +44,7 @@ jobs:
         run: |
           libtorch/build_docker.sh
   build-docker-rocm:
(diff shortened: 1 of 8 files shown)


jithunnair-amd:
ROCm manywheel/libtorch docker images built successfully. Checked that:
* ROCm was not re-installed via install_rocm.sh
* Magma was built from source
* MIOpen was built from source

atalman (approved):
lgtm

PR #1427, opened 2023-06-13
[Manywheel] Add Python-3.12.0b2

malfet:
To enable initial experiments with PyTorch builds

### DIFF
--- a/common/install_cpython.sh
+++ b/common/install_cpython.sh
@@ -5,7 +5,7 @@ PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
 GET_PIP_URL=https://bootstrap.pypa.io/get-pip.py
 
 # Python versions to be installed in /opt/$VERSION_NO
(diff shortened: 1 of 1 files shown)


common/install_cpython.sh
```diff
@@ -5,7 +5,7 @@ PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
 GET_PIP_URL=https://bootstrap.pypa.io/get-pip.py
 
 # Python versions to be installed in /opt/$VERSION_NO
-CPYTHON_VERSIONS=${CPYTHON_VERSIONS:-"3.7.5 3.8.1 3.9.0 3.10.1 3.11.0"}
+CPYTHON_VERSIONS=${CPYTHON_VERSIONS:-"3.7.5 3.8.1 3.9.0 3.10.1 3.11.0 3.12.0"}
```
albanD, line 8: unrelated to this change but according to our policy, If these represent the lowest patch version we support for each, we can bump the patch versions all the way to the first non-bugfix version. It doesn't change much usually but I wonder how strict we want to be to our rule?

PR #1433, opened 2023-06-17
Provide file hashes in the URLs to avoid unnecessary file downloads (bandwidth saver)

### LINKED ISSUE #1347: Pip packaging and publishing improvements in pytorch wheels for better integration with poetry
Please see the issues described here: https://github.com/pytorch/pytorch/issues/76557

matteius:
Supply sha256 query parameters using boto3 to avoid hundreds of extra Gigabytes of downloads each day during pipenv and poetry resolution lock cycles.

Fixes point 1 in https://github.com/pytorch/pytorch/issues/76557
Fixes #1347

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -3,6 +3,7 @@
 import argparse
 import base64
 import dataclasses
+import functools
 import time
 
 from os import path, makedirs
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -281,9 +281,7 @@ class S3Index:
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
         for obj in sorted(self.gen_file_list(subdir, package_name)):
-            maybe_fragment = ""
-            if obj.checksum:
-                maybe_fragment = f"#sha256={obj.checksum}"
+            maybe_fragment = f"#sha256={obj.checksum}" if obj.checksum else ""
             out.append(f'    <a href="/{obj}{maybe_fragment}">{path.basename(obj).replace("%2B","+")}</a><br/>')
         # Adding html footer
         out.append('  </body>')
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -11,11 +11,9 @@ from re import sub, match, search
 from packaging.version import parse
 
 import boto3
-from botocore.exceptions import NoCredentialsError
 
 
 S3 = boto3.resource('s3')
-CLIENT = boto3.client('s3')
 BUCKET = S3.Bucket('pytorch')
 
 ACCEPTED_FILE_EXTENSIONS = ("whl", "zip", "tar.gz")
@@ -122,8 +120,8 @@ def between_bad_dates(package_build_time: datetime):
 
 
 class S3Index:
-    def __init__(self: S3IndexType, objects: List[str], prefix: str) -> None:
-        self.objects = objects
+    def __init__(self: S3IndexType, objects: Dict[str, str], prefix: str) -> None:
+        self.objects = objects  # s3 key to checksum mapping
         self.prefix = prefix.rstrip("/")
         self.html_name = PREFIXES_WITH_HTML[self.prefix]
         # should dynamically grab subdirectories like whl/test/cu101
@@ -147,7 +145,7 @@ class S3Index:
         # also includes versions without GPU specifier (i.e. cu102) for easier
         # sorting, sorts in reverse to put the most recent versions first
         all_sorted_packages = sorted(
-            {self.normalize_package_version(obj) for obj in self.objects},
+            {self.normalize_package_version(s3_key) for s3_key in self.objects.keys()},
             key=lambda name_ver: parse(name_ver.split('-', 1)[-1]),
             reverse=True,
         )
@@ -167,10 +165,12 @@ class S3Index:
                 to_hide.add(obj)
             else:
                 packages[package_name] += 1
-        return set(self.objects).difference({
-            obj for obj in self.objects
-            if self.normalize_package_version(obj) in to_hide
-        })
+        nightly_packages = {}
+        for obj, checksum in self.objects.items():
+            normalized_package_version = self.normalize_package_version(obj)
+            if not normalized_package_version in to_hide:
+                nightly_packages[normalized_package_version] = checksum
+        return nightly_packages
 
     def is_obj_at_root(self, obj:str) -> bool:
         return path.dirname(obj) == self.prefix
@@ -191,15 +191,15 @@ class S3Index:
             else self.objects
         )
         subdir = self._resolve_subdir(subdir) + '/'
-        for obj in objects:
+        for obj, checksum in objects.items():
             if package_name is not None:
                 if self.obj_to_package_name(obj) != package_name:
                     continue
             if self.is_obj_at_root(obj) or obj.startswith(subdir):
-                yield obj
+                yield obj, checksum
 
     def get_package_names(self, subdir: Optional[str] = None) -> List[str]:
-        return sorted(set(self.obj_to_package_name(obj) for obj in self.gen_file_list(subdir)))
+        return sorted(set(self.obj_to_package_name(obj) for obj, _ in self.gen_file_list(subdir)))
 
     def normalize_package_version(self: S3IndexType, obj: str) -> str:
         # removes the GPU specifier from the package name as well as
@@ -212,23 +212,6 @@ class S3Index:
 
     def obj_to_package_name(self, obj: str) -> str:
         return path.basename(obj).split('-', 1)[0]
-
-    def fetch_checksum_from_s3(self, s3_key):
-        s3_key = s3_key.replace("%2B", "+")
-        try:
-            response = CLIENT.get_object_attributes(
-                Bucket=BUCKET,
-                Key=s3_key,
-                ObjectAttributes=['Checksum']
-            )
-            checksum = response['Checksum']['ChecksumSHA256']
-            return checksum
-        except NoCredentialsError:
-            print("No AWS credentials found")
-            return None
-        except Exception as e:
-            print(f"Unable to retrieve checksum due to {e}")
-            return None
 
     def to_legacy_html(
         self,
@@ -244,7 +227,7 @@ class S3Index:
         out: List[str] = []
         subdir = self._resolve_subdir(subdir)
         is_root = subdir == self.prefix
-        for obj in self.gen_file_list(subdir):
+        for obj, _ in self.gen_file_list(subdir):
             # Strip our prefix
             sanitized_obj = obj.replace(subdir, "", 1)
             if sanitized_obj.startswith('/'):
@@ -272,8 +255,7 @@ class S3Index:
         out.append('<html>')
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
-        for obj in sorted(self.gen_file_list(subdir, package_name)):
-            checksum = self.fetch_checksum_from_s3(obj)
+        for obj, checksum in sorted(self.gen_file_list(subdir, package_name)):
             if checksum:
                 out.append(f'    <a href="/{obj}#sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
             else:
@@ -338,7 +320,6 @@ class S3Index:
                     Body=self.to_simple_package_html(subdir=subdir, package_name=pkg_name)
                 )
 
-
     def save_legacy_html(self) -> None:
         for subdir in self.subdirs:
             print(f"INFO Saving {subdir}/{self.html_name}")
@@ -370,9 +351,12 @@ class S3Index:
                 for pattern in ACCEPTED_SUBDIR_PATTERNS
             ]) and obj.key.endswith(ACCEPTED_FILE_EXTENSIONS)
             if is_acceptable:
+                response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
+                sha256 = response.get("ChecksumSHA256")
                 sanitized_key = obj.key.replace("+", "%2B")
-                objects.append(sanitized_key)
+                objects.append((sanitized_key, sha256))
         return cls(objects, prefix)
+
 
 def create_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser("Manage S3 HTML indices for PyTorch")
@@ -384,6 +368,7 @@ def create_parser() -> argparse.ArgumentParser:
     parser.add_argument("--do-not-upload", action="store_true")
     parser.add_argument("--generate-pep503", action="store_true")
     return parser
+
 
 def main():
     parser = create_parser()
@@ -408,6 +393,7 @@ def main():
             idx.upload_legacy_html()
             if args.generate_pep503:
                 idx.upload_pep503_htmls()
+
 
 if __name__ == "__main__":
     main()
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -274,7 +274,10 @@ class S3Index:
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
         for obj in sorted(self.gen_file_list(subdir, package_name)):
             checksum = self.fetch_checksum_from_s3(obj)
-            out.append(f'    <a href="/{obj}?sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+            if checksum:
+                out.append(f'    <a href="/{obj}?sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+            else:
+                out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
         # Adding html footer
         out.append('  </body>')
         out.append('</html>')
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -275,7 +275,7 @@ class S3Index:
         for obj in sorted(self.gen_file_list(subdir, package_name)):
             checksum = self.fetch_checksum_from_s3(obj)
             if checksum:
-                out.append(f'    <a href="/{obj}?sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+                out.append(f'    <a href="/{obj}#sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
             else:
                 out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
         # Adding html footer
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -11,6 +11,7 @@ from re import sub, match, search
 from packaging.version import parse
 
 import boto3
+from botocore.exceptions import NoCredentialsError
 
 
 S3 = boto3.resource('s3')
@@ -212,6 +213,23 @@ class S3Index:
     def obj_to_package_name(self, obj: str) -> str:
         return path.basename(obj).split('-', 1)[0]
 
+    def fetch_checksum_from_s3(self, s3_key):
+        s3_key = s3_key.replace("%2B", "+")
+        try:
+            response = CLIENT.get_object_attributes(
+                Bucket=BUCKET,
+                Key=s3_key,
+                ObjectAttributes=['Checksum']
+            )
+            checksum = response['Checksum']['ChecksumSHA256']
+            return checksum
+        except NoCredentialsError:
+            print("No AWS credentials found")
+            return None
+        except Exception as e:
+            print(f"Unable to retrieve checksum due to {e}")
+            return None
+
     def to_legacy_html(
         self,
         subdir: Optional[str]=None
@@ -255,7 +273,8 @@ class S3Index:
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
         for obj in sorted(self.gen_file_list(subdir, package_name)):
-            out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+            checksum = self.fetch_checksum_from_s3(obj)
+            out.append(f'    <a href="/{obj}?sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
         # Adding html footer
         out.append('  </body>')
         out.append('</html>')
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -339,7 +339,7 @@ class S3Index:
 
     @classmethod
     def from_S3(cls: Type[S3IndexType], prefix: str) -> S3IndexType:
-        objects = []
+        objects = {}
         prefix = prefix.rstrip("/")
         for obj in BUCKET.objects.filter(Prefix=prefix):
             is_acceptable = any([path.dirname(obj.key) == prefix] + [
@@ -354,7 +354,7 @@ class S3Index:
                 response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
                 sha256 = response.get("ChecksumSHA256")
                 sanitized_key = obj.key.replace("+", "%2B")
-                objects.append((sanitized_key, sha256))
+                objects[sanitized_key] = sha256
         return cls(objects, prefix)
 
 
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -255,10 +255,10 @@ class S3Index:
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
         for obj, checksum in sorted(self.gen_file_list(subdir, package_name)):
+            maybe_fragment = ""
             if checksum:
-                out.append(f'    <a href="/{obj}#sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
-            else:
-                out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+                maybe_fragment = f"#sha256={checksum}"
+            out.append(f'    <a href="/{obj}{maybe_fragment}">{path.basename(obj).replace("%2B","+")}</a><br/>')
         # Adding html footer
         out.append('  </body>')
         out.append('</html>')
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -107,6 +107,7 @@ S3IndexType = TypeVar('S3IndexType', bound='S3Index')
 
 
 @dataclasses.dataclass(frozen=True)
+@functools.total_ordering
 class S3Object:
     key: str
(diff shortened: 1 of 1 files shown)


matteius:
I just signed the Contributor License Agreement,

matteius:
@malfet Could I get this on your radar?

matteius:
@seemethere or @atalman -- Could I get this PR on your radar?

matteius:
@thejcannon I was only tackling  the first item on the list -- which is going to be mutually beneficial for pipenv.  I am not that familiar with PEP 658 and that should be pursued in a separate PR.  In the case of pipenv, this change alone should allow us to go from downloading all of the matching distributions to just the one relevant to the system (at least I think it will).  Checking on the # in the URL now ..

thejcannon:
Gotcha. I started on the other bullet points and have a branch.

Two questions for you:
- I haven't used pipenv, can you help my understanding of why the hashes of wheels help? Is the assumption that the index could theoretically serve a different binary blob for a given package+version at some later point in time? Therefore without a hash pipenv (and others) re-download to be the most correct?
- How did you test the S3 checksum grabbing? When I issue a similar command (`aws s3api get-object-attributes ...`) I get denied permissions :disappointed:

matteius:
>  can you help my understanding of why the hashes of wheels help? 

@thejcannon  Yes, the issue is consider this lock file entry for pytorch:
```
        "torch": {
            "hashes": [
                "sha256:4b4bf40d844fdcf9b7261e62aff9eede7c89c2b904343b3f9dc12393763779cb",
                "sha256:566cbfad47b25ef6b1023518982179439d8e53069f3d2341e8adcd1e4b2ef179",
                "sha256:588188a37e2f2c9bb8f807e0d49ac4d66b9ef23f4417233247c5d2f796c4f42d",
                "sha256:78ae2260ac1902c5a224fe1b7994072bbfa7ef8a1734d8697ca74249fa6391c5",
                "sha256:7c23faff945e98c6e242ce41a480e7fe48426140d5994ea3b9886c02145d58cf",
                "sha256:bbe445308a43960efeabc4f4a4b6c5d0db209ef5d5148672c750943d1ea42f54",
                "sha256:bcb611272d7aa5c2f7c1648b2227303a023b1de46aabccf1846a8aca1bfb5cd8",
                "sha256:ccbd72f39d76fb7e7db6513bcc8e4ee684348f9865bd8cddca6ce2dd07b3ccd0"
            ],
            "index": "downloadpytorch",
            "version": "==1.10.2+cu111"
        },
```

To generate this proper lock file entry the lock phase had to download every single wheel to get the hash value from the registry, but with the above change (assuming the AWS ACL is correct and I didn't make a mistake) it will have the hashes and scrape them from the html page instead.

The hashes are used during the install phase to verify that the package downloaded matches what was determined when the lock file was was locked or last updated.  So you can see where if each of those wheels is ~1.5 GB that is ~12 GB of downloads that would be avoided during lock phase.   The idea is that all hashes related to that version are included because the lock file is shared across system platforms so the install phase may use different package+hash based on the system but everything in that list was authorized.

> How did you test the S3 checksum grabbing?

I didn't really, I don't have anything in AWS but am familiar with using it and boto3, and wrote that code just from reading how to get the hash in documentation, but its possible I could have made a mistake: 
 https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_object_attributes.html

matteius:
Hmm, I guess those docs also say:
> ChecksumSHA256 (string) –
The base64-encoded, 256-bit SHA-256 digest of the object. This will only be present if it was uploaded with the object. 

That seems possibly troublesome.

matteius:
image

Which links to:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html



image
https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/

It seems that:
1.) The bucket needs to be configured to use the sha-256 checksum (md5 is default)
2.) Historical files probably wouldn't get this by default without being "copied" after the setting is changed, possibly also explicitly specifying the sha-256 algorithm in the copy or put_object commands.

thejcannon:
Ah ok your explanation makes things crystal clear now, thank you.

---

I also reached out to some folks at Meta to help me with the S3 side of PEP 658, we can also see if checksum enabling (including historical files) is something their willing to enable.

thejcannon:
FYI @matteius 

- I think you should probably tag this PR as fixes https://github.com/pytorch/builder/issues/1347 :stuck_out_tongue: 
- I got in contact with some folks @malfet @seemethere who are going to look into adding the checksums to the objects in S3 (currently the `Checksum` is not an attribute attached to the objects, so your S3 command will always fail)
- Depending on the timing of the merges and checksums, please consider updating this PR with support for metadata files as well: https://github.com/pytorch/builder/pull/1457 :heart:

s3_management/manage.py
```diff
@@ -212,6 +213,23 @@ def normalize_package_version(self: S3IndexType, obj: str) -> str:
...
+            response = CLIENT.get_object_attributes(
+                Bucket=BUCKET,
+                Key=s3_key,
+                ObjectAttributes=['Checksum']
+            )
+            checksum = response['Checksum']['ChecksumSHA256']
+            return checksum
+        except NoCredentialsError:
+            print("No AWS credentials found")
+            return None
```
thejcannon, line 228: Why not just let this propogate? Otherwise it'll just repeatedly print the error and generate a sub-optimal index.

s3_management/manage.py
```diff
@@ -212,6 +213,23 @@ def normalize_package_version(self: S3IndexType, obj: str) -> str:
...
+        except Exception as e:
```
thejcannon, line 229: This is a wide exception. I think the `try:` block and relevant exceptions likely should be scoped to be explicit about what is expected to fail and what isn't.

Otherwise, a simple bug could manifest as an index not having hashes. When the alternative is an error with a stacktrace given to the user, and no index generated.

s3_management/manage.py
```diff
@@ -212,6 +213,23 @@ def normalize_package_version(self: S3IndexType, obj: str) -> str:
...
 
+    def fetch_checksum_from_s3(self, s3_key):
+        s3_key = s3_key.replace("%2B", "+")
+        try:
+            response = CLIENT.get_object_attributes(
+                Bucket=BUCKET,
+                Key=s3_key,
+                ObjectAttributes=['Checksum']
+            )
+            checksum = response['Checksum']['ChecksumSHA256']
```
thejcannon, line 224: Is there any way to leverage the call in `from_S3` to also include this information?

If not, perhaps this should be done there so the returned objects already have the checksum attached and the downstream code can choose to use it or not?

That would isolate the S3 retrieval code to be in one logical place.

thejcannon (commented):
While I'm around I thought I'd review the code :yum: 

Exciting stuff!

s3_management/manage.py
matteius: My thinking was that the index page should still generate as it does now even if there is no hash available to provide.   I should definitely remove the print statements.  I am not sure what exceptions would be raised before the `Checksum` values get added to the objects.

thejcannon: (Did this mean to get commented below? on https://github.com/pytorch/builder/pull/1433/files/b3799a8eb42b56118a4bc0658902875734888721#r1261472864? If not I'm not following the thread here. If the credentials aren't found we really can't know if there's a checksum or not)

s3_management/manage.py
matteius: > Is there any way to leverage the call in from_S3 to also include this information?

From what I can tell they are different clients/endpoints that serve different purposes.

>If not, perhaps this should be done there so the returned objects already have the checksum attached and the downstream code can choose to use it or not?

From my perspective, the ``from_S3` appears to just be tracking a list of strings which is the object s3_key with `List[str]` so I am not sure it makes sense to tackle a larger refactor just to add all the additional calls to pre-lookup these checksums and store them in memory until their point of use.  

> That would isolate the S3 retrieval code to be in one logical place.

Isn't it already isolated to this one file manage.py where the clients are in scope for the whole file?

Maybe I am being dense and don't see the obvious way to accomplish this without spending a couple more hours on it, plus the more complexity I add, well I have no real way of testing it.  Looking for any guidance on the right exceptions to catch or not catch above, but I think we need to catch something to cover the case that s3 isn't ready or able to provide the hashes.

s3_management/manage.py
matteius: I mean its the same code block in my mind -- I am just saying I don't know what exceptions we want to catch and handle for the case of the checksum isn't available yet -- maybe we don't care about `NoCredentialsError` but I think we need to handle _something_

thejcannon: I think it's safe to assume:
- The runner has valid credentials to the object
- The runner has the correct permissions to access the attributes
- S3 is alive and well and will return good JSON.

Because otherwise if these assumptions aren't held, there's a chance the script generates an `index.html` without the checksums and we're back to square 1.

So I think the only thing to handle here is "this object has no checksum attribute".

Which means the code looks roughly like:
```python
response = CLIENT.get_object_attributes(
    Bucket=BUCKET,
    Key=s3_key,
    ObjectAttributes=["Checksum"]
)
checksum = response.get("Checksum", {}).get("ChecksumSHA256", None)
```

without any `try`/`except`.

FWIW I tested your code against a work bucket and got `TypeError: expected string or bytes-like object`. Switching the argument for `Bucket` to the string bucket name then gives me a valid response.

So probably use `BUCKET.name` as well.

---

Here's some local testing...

```python
>>> import boto3
>>> BUCKET=S3.Bucket("<bucketname>")
>>> CLIENT = boto3.client("s3")
>>> response = CLIENT.get_object_attributes(Bucket=BUCKET.name,Key="<object-without-checksum>",ObjectAttributes=['Checksum'])
>>> list(response)
['ResponseMetadata', 'LastModified', 'VersionId']
>>> response.get("Checksum", {}).get("ChecksumSHA256", None)
>>> response = CLIENT.get_object_attributes(Bucket=BUCKET.name,Key="<object-WITH-checksum>",ObjectAttributes=['Checksum'])
>>> list(response)
['ResponseMetadata', 'LastModified', 'Checksum']
>>> response.get("Checksum", {}).get("ChecksumSHA256", None)
'7Uc3AtM1Dlc4XiQ8A6NkQdV9JMU4liPtz7wOd+RRCTE='
```

s3_management/manage.py
thejcannon: OK, so looks like you should be able to do this in `from_s3` (which is ideal, because we'll want the checksums for metadata files in https://github.com/pytorch/builder/pull/1457 as well:

```python
>>> import boto3
>>> BUCKET=S3.Bucket("<bucketname>")
>>> obj = next(iter(BUCKET.objects.filter(Prefix="<some prefix>")))
>>> response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
>>> response["ChecksumSHA256"]
'7Uc3AtM1Dlc4XiQ8A6NkQdV9JMU4liPtz7wOd+RRCTE='
```

And then for an object without a checksum:
```python
>>> >>> response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
>>> list(response)
['ResponseMetadata', 'AcceptRanges', 'LastModified', 'ContentLength', 'ETag', 'ContentType', 'ServerSideEncryption', 'Metadata']
```

thejcannon:
If you don't want to put more man-hours into this, I wouldn't mind picking this up to be done alongside/after https://github.com/pytorch/builder/pull/1457.

I want the same bandwidth savings for myself, my engineers, and everyone else. So I'm really eager to get this in! :smile:

matteius:
> If you don't want to put more man-hours into this

I'd like to see it through, thanks for your guidance -- I'll push some more commits soon.

thejcannon:
Bitchin' And thanks for your patience. Hopefully by the time the Meta folks stop by this'll be an easy approval :+1:

matteius:
@thejcannon This is how I would track the checksums, could you have a look and see if it makes sense to you too?

s3_management/manage.py
```diff
@@ -226,7 +227,7 @@ def to_legacy_html(
         out: List[str] = []
         subdir = self._resolve_subdir(subdir)
         is_root = subdir == self.prefix
-        for obj in self.gen_file_list(subdir):
+        for obj, _ in self.gen_file_list(subdir):
```
matteius, line 230: I am not as familiar with the legacy html, but perhaps it too should have the checksums?

s3_management/manage.py
```diff
@@ -348,10 +351,13 @@ def from_S3(cls: Type[S3IndexType], prefix: str) -> S3IndexType:
                 for pattern in ACCEPTED_SUBDIR_PATTERNS
             ]) and obj.key.endswith(ACCEPTED_FILE_EXTENSIONS)
             if is_acceptable:
+                response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
+                sha256 = response.get("ChecksumSHA256")
                 sanitized_key = obj.key.replace("+", "%2B")
-                objects.append(sanitized_key)
+                objects.append((sanitized_key, sha256))
```
thejcannon, line 357: This still has `objects` as a list. I don't see it being converted to a `dict` anywhere?

I think you can just change it on line 343 to a `dict` and here to `objects[sanitized_key] = sha256`

s3_management/manage.py
```diff
@@ -254,8 +255,11 @@ def to_simple_package_html(
         out.append('<html>')
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
-        for obj in sorted(self.gen_file_list(subdir, package_name)):
-            out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+        for obj, checksum in sorted(self.gen_file_list(subdir, package_name)):
+            if checksum:
+                out.append(f'    <a href="/{obj}#sha256={checksum}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+            else:
+                out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
```
thejcannon, line 262:
```suggestion
            maybe_fragment = ""
            if checksum:
                maybe_fragment = f"#sha256={checksum}""
            out.append(f'    <a href="/{obj}{maybe_fragment}">{path.basename(obj).replace("%2B","+")}</a><br/>')
```

s3_management/manage.py
```diff
@@ -121,8 +120,8 @@ def between_bad_dates(package_build_time: datetime):
 
 
 class S3Index:
-    def __init__(self: S3IndexType, objects: List[str], prefix: str) -> None:
-        self.objects = objects
+    def __init__(self: S3IndexType, objects: Dict[str, str], prefix: str) -> None:
```
thejcannon, line 123: The RHS should be `str | None`

```suggestion
    def __init__(self: S3IndexType, objects: Dict[str, str | None], prefix: str) -> None:
```

s3_management/manage.py
```diff
@@ -146,7 +145,7 @@ def nightly_packages_to_show(self: S3IndexType) -> Set[str]:
         # also includes versions without GPU specifier (i.e. cu102) for easier
         # sorting, sorts in reverse to put the most recent versions first
         all_sorted_packages = sorted(
-            {self.normalize_package_version(obj) for obj in self.objects},
+            {self.normalize_package_version(s3_key) for s3_key in self.objects.keys()},
```
thejcannon, line 148: `flake8-simplify` ([link](https://pypi.org/project/flake8-simplify/)) suggests just iterating over the dict instead of explicitly calling `.keys()`. We could talk about the likely very itty bitty performance bump, but lol.

Is there a reason you prefer `.keys()`?

s3_management/manage.py
```diff
@@ -166,10 +165,12 @@ def nightly_packages_to_show(self: S3IndexType) -> Set[str]:
...
-        return set(self.objects).difference({
-            obj for obj in self.objects
-            if self.normalize_package_version(obj) in to_hide
-        })
+        nightly_packages = {}
+        for obj, checksum in self.objects.items():
+            normalized_package_version = self.normalize_package_version(obj)
+            if not normalized_package_version in to_hide:
+                nightly_packages[normalized_package_version] = checksum
+        return nightly_packages
```
thejcannon, line 173: Maybe simplify to a comprehension?

```suggestion
        return {
            s3_key: checksum
            for s3_key, checksum in self.objects.items()
            if self.normalize_package_version(s3_key) not in to_hide
        }
```

s3_management/manage.py
```diff
@@ -190,15 +191,15 @@ def gen_file_list(
...
         )
         subdir = self._resolve_subdir(subdir) + '/'
-        for obj in objects:
+        for obj, checksum in objects.items():
             if package_name is not None:
                 if self.obj_to_package_name(obj) != package_name:
                     continue
             if self.is_obj_at_root(obj) or obj.startswith(subdir):
-                yield obj
+                yield obj, checksum
```
thejcannon, line 199: The return type of this function needs updating

s3_management/manage.py
thejcannon, line 173: Also the return type needs updating

s3_management/manage.py
```diff
@@ -348,10 +351,13 @@ def from_S3(cls: Type[S3IndexType], prefix: str) -> S3IndexType:
                 for pattern in ACCEPTED_SUBDIR_PATTERNS
             ]) and obj.key.endswith(ACCEPTED_FILE_EXTENSIONS)
             if is_acceptable:
+                response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
+                sha256 = response.get("ChecksumSHA256")
```
thejcannon, line 355: Probably worth a comment explaining why we go through the hassle here.

Something like:
`# Add PEP 503-compatible hashes to URLs to allow clients to avoid re-downloads, if possible.

thejcannon (commented):
Looking really good. Likely the last wave of comments :+1:

s3_management/manage.py
thejcannon: Resolved.

s3_management/manage.py
thejcannon: Resolved.

s3_management/manage.py
thejcannon: I think there's a couple other places this should be done as well.

s3_management/manage.py
matteius: Wouldn't it need to be this which would duplicate the call to `normalize_package_version`?
```
        return {
            self.normalize_package_version(s3_key): checksum
            for s3_key, checksum in self.objects.items()
            if self.normalize_package_version(s3_key) not in to_hide
        }
```

thejcannon: As best I can tell, the old code did `set(self.objects)` meaning it wasn't using the return of `normalize_package_version` as the key.

s3_management/manage.py
matteius: I believe this is fixed now.

matteius:
If we could get some stats from Meta about the S3 cost savings (did some napkin math, its probably way wrong, but I estimate could be north of 200,000$ a month savings if `200,619*16*0.08` is a meaningful calculation (the average monthly downloads from pypi torch, but I assume the gpu specific indexes are more popular) -- anyway would be interesting to have insight into 😁

vit-zikmund:
Hey @matteius & @thejcannon , I just wanted to say thanks for trying to push all this further. All I had time for so far was the [analysis in #1347](https://github.com/pytorch/builder/issues/1347#issuecomment-1576683075), I hope it was at least somewhat helpful :smile: 
While I'm not in any particular position for suggesting anything for the PR here, here's something that should hopefully make the change thinner and a bit better to manage.
 While the `obj[ect]` in the code is just a `str` with a local path (or an s3 key), you could define a `dataclass` to represent the object with more attributes like this:
```python
@dataclasses.dataclass(frozen=True)
class S3Object:
  key: str
  checksum: str | None

def __str__(self):
  return self.key
```
 By overriding the `__str__` method it gets to be printed as before. This might not be enough in some places and one will need to reach for the `obj.key`, but that shouldn't be a big hassle. This is extensible beyond just the `checksum` and one can keep all the previous sets & lists you switched for a dict. This dataclass might also bear a `metadata: str | bool | None` property, which could hold the metadata presence boolean (or better its checksum) for @thejcannon's [PEP 658 PR](https://github.com/pytorch/builder/pull/1457/files#diff-b4c713cb4f974318a824c8d87db6e78eec6b4b9e129b3e5a83f4268bfb3320b0).

thejcannon:
Oh yeah. I LOVE me some dataclasses ([no seriously](https://peps.python.org/pep-0712/)). I think there's obviously some trepidation in refactoring code you don't-quite-own, so that kind of change is likely a good idea to have in a dedicated PR and perhaps encouraged to be merged before this or https://github.com/pytorch/builder/pull/1457/files#diff-b4c713cb4f974318a824c8d87db6e78eec6b4b9e129b3e5a83f4268bfb3320b0.

s3_management/manage.py
```diff
@@ -348,10 +350,14 @@ def from_S3(cls: Type[S3IndexType], prefix: str) -> S3IndexType:
                 for pattern in ACCEPTED_SUBDIR_PATTERNS
             ]) and obj.key.endswith(ACCEPTED_FILE_EXTENSIONS)
             if is_acceptable:
+                # Add PEP 503-compatible hashes to URLs to allow clients to avoid spurious downloads, if possible.
+                response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
+                sha256 = response.get("ChecksumSHA256")
```
vit-zikmund, line 355: Hi there @matteius, I believe this needs a little bit more polishing. The s3 checksum seems to be a base64 encoded byte value of the hash (judging by what's to be seen [here](https://github.com/pytorch/builder/pull/1433#discussion_r1261876922) or [here](https://github.com/aws/aws-cli/issues/6750#issuecomment-1619191575)), however what is required by the PEP 503 is a hex encoded value.
```suggestion
                response = obj.meta.client.head_object(Bucket=BUCKET.name, Key=obj.key, ChecksumMode="ENABLED")
                sha256 = (_b64 := response.get("ChecksumSHA256")) and base64.b64decode(_b64).hex()
```
This will also need an `import base64` at the top.

vit-zikmund:
> that kind of change is likely a good idea to have in a dedicated PR and perhaps encouraged to be merged before this or https://github.com/pytorch/builder/pull/1457/files#diff-b4c713cb4f974318a824c8d87db6e78eec6b4b9e129b3e5a83f4268bfb3320b0.
Thanks for your kind words @thejcannon. I don't want to hijack this thread more than I feel necessary (crossed that already), but with all due respect I don't see any reason in pushing such a standalone refactoring change in my own PR given there's been over a month of zero activity on #1347 after I tagged 2 owners (hopefully politely enough). I read you write you got in contact with them, so I hope there's some real collaboration going on behind what we see here. I never got that far and thus will crawl back to my lair and wait for you to play the big game to the successful end, wishing you (us) all the best.

Either way, you both guys have a partially conflicting & related PRs now, so if luck has it one will need to rebase & resolve one's PR on top of the other's. My dataclass suggestion that can tie these two together lies in the field to be picked up or not, no problem. Let's focus on getting anything that makes sense to work and stop all these agonizing downloads. Thank you! :crossed_fingers:

s3_management/manage.py
thejcannon: Very good eye! :+1: (The first clue should've been the length :joy:)

thejcannon:
> I read you write you got in contact with them, so I hope there's some real collaboration going on behind what we see here. I never got that far and thus will crawl back to my lair and wait for you to play the big game to the successful end, wishing you (us) all the best.

I wouldn't be so quick to disparage yourself in a community of peers. I only got in contact with some folks by nagging a very nice Meta employee I met at PyCon earlier this year. Then they suggested we have the discussion(s) out in the open, so also nothing going on behind the scenes. As much excitement as we all have I'm sure the slow wheel of Enterprise is churning on, being methodical and deliberate. So now we just wait I suppose (until I feel the need to nag some more :wink: )

Your thorough comment on https://github.com/pytorch/builder/issues/1347 and your eye for detail [above](https://github.com/pytorch/builder/pull/1433#discussion_r1263936261) only proves your input in these discussions is valuable. So I would happily review (and outsiders-approve) a PR you put forth for your great idea. Then it'd just be a race to see who gets accepted first :smiling_imp:

matteius:
I took a shot at the refactor to use the dataclass.

s3_management/manage.py
```diff
@@ -104,6 +105,31 @@
...
+        return self.key < other.key
+
+    def __le__(self, other):
+        return self.key <= other.key
+
+    def __gt__(self, other):
+        return self.key > other.key
+
+    def __ge__(self, other):
+        return self.key >= other.key
```
vit-zikmund, line 130: We can squeeze this with the `@functools.total_ordering` decorator into 2 methods. Also I believe you meant to define `__eq__` instead of `__cmp__` which is a python2-only thing.
```suggestion
@dataclasses.dataclass(frozen=True)
@functools.total_ordering
class S3Object:
    key: str
    checksum: str | None

    def __str__(self):
        return self.key

    def __eq__(self, other):
        return self.key == other.key

    def __lt__(self, other):
        return self.key < other.key
```

s3_management/manage.py
```diff
@@ -1,20 +1,21 @@
 #!/usr/bin/env python
 
 import argparse
+import base64
+import dataclasses
```
vit-zikmund, line 5: This is a followup to the other comment.
```suggestion
import dataclasses
import functools
```

vit-zikmund (commented):
Hey @matteius, I'm really honored you gave my suggestion a chance! :bow: It looks great. The resulting diff is indeed smaller and better to read :wink: 

I've took the liberty to check the result and drop a few comments/suggestions, most of which try to make the new code occupy a tiny bit less lines.

Thank you once again!

vit-zikmund (approved):
Nice :+1:  Now to get some real stakeholders to notice :hourglass:

malfet (approved):

thejcannon:
@matteius congrats

@malfet wanna review the accompanying PR? https://github.com/pytorch/builder/pull/1457

malfet:
@thejcannon first I need to fix all the bugs introduced by this change (first of them, is that `str | None` is not a valid type annotation), but sure, will try to review this one soon

thejcannon:
`from __future__ import annotations`?

Also consider that these kinds of bugs are introduced by not having sufficient checks gate the PR

s3_management/manage.py
matteius: Build is failing with:
```botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in input: "ChecksumMode", must be one of: Bucket, IfMatch, IfModifiedSince, IfNoneMatch, IfUnmodifiedSince, Key, Range, VersionId, SSECustomerAlgorithm, SSECustomerKey, SSECustomerKeyMD5, RequestPayer, PartNumber```

matteius: https://github.com/pytorch/builder/pull/1548

vit-zikmund: Tagging @malfet again, as these inner comments are notoriously hard to discover. Hope it's no big deal 😇

PR #1438, opened 2023-06-23
Update build CD for aarch64 Wheel

xncqr:
This is an update to the aarch64 Linux Wheel build for PyTorch CD.

Updates:
* Removed builds of Vision, Audio, Text, and Data as they need to be made on their perspective repos.
* Created separate def for completing the wheel with auditwheel and putting into the artifacts folder.
* Removed git install as it was already installed.
* Added libgomp OS install so it is no longer needed to be injected as auditwheel handles it.
* Added MKLDNN check for aarch64 over standard MKL as aarch64 is not "Intel" based.

This build was tested at https://github.com/xncqr/pytorch/actions/runs/5351891068 with successful build and test of aarch64 wheel build.

### DIFF
--- a/aarch64_linux/aarch64_ci_build.sh
+++ b/aarch64_linux/aarch64_ci_build.sh
@@ -12,7 +12,7 @@ PATH=/opt/conda/bin:$PATH
 # Install OS dependent packages
 ###############################################################################
 yum -y install epel-release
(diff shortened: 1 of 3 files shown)


xncqr:
@seemethere, @malfet, or @atalman,

PR is ready for your review with information in description.

PR #1439, opened 2023-06-29
Run make_wheel_record parallel in background

xwang233:
By running `make_wheel_record` parallel in background, this saves ~8 minutes on my 12-core intel machine with a full cuda wheel build.

https://unix.stackexchange.com/questions/42544/does-redirecting-output-to-a-file-apply-a-lock-on-the-file/42564#42564
This answer suggested that each line will still be a line will written in parallel to the record file.

https://unix.stackexchange.com/questions/103920/parallelize-a-bash-for-loop/436713#436713
If overloading CPU is a concern, we may try this native `wait -n` bash trick to limit the concurrency.


xwang233:
cc @malfet @atalman @ptrblck

atalman:
@xwang233 please open test PR in pytorch/pytorch for this change and attach ciflow/binaries label to it so that it will actually trigger binary builds

PR #1448, opened 2023-07-06
Update Docker base images for conda and libtorch

malfet:
Followup after https://github.com/pytorch/builder/pull/1446

CUDA-10.2 and moreover CUDA-9.2 docker images are gone per [Nvidia's Container Support Policy](https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/support-policy.md):
> After a period of Six Months time, the EOL tags WILL BE DELETED from Docker Hub and Nvidia GPU Cloud (NGC). This deletion ensures unsupported tags (and image layers) are not left lying around for customers to continue using after they have long been abandoned.

Also, as all our Docker script install CUDA toolkit anyway, what's the point of using `nvidia/cuda` images at all instead of `centos:7`/`ubuntu:18.04` that former are based on, according to https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.4.3/centos7/base/Dockerfile

Explicitly install `g++` to `libtorch/Docker` base image, as it's needed by `patchelf`

Please note, that `libtorch/Docker` can not be completed without buildkit, as `rocm` step depends on `python3` which is not available in `cpu` image

### DIFF
--- a/conda/Dockerfile
+++ b/conda/Dockerfile
@@ -1,6 +1,6 @@
 ARG CUDA_VERSION=10.2
 ARG BASE_TARGET=cuda${CUDA_VERSION}
-FROM nvidia/cuda:9.2-devel-centos7 as base
+FROM nvidia/cuda:11.4.3-devel-centos7 as base
 
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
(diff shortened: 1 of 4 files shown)


atalman (approved):
LGTM

PR #1451, opened 2023-07-07
Rebuild docker images on release builds

atalman:
Rebuild docker images on release builds. It should also tag images for release here: https://github.com/pytorch/builder/blob/3fc310ac21c9ede8d0ce13ec71096820a41eb9f4/conda/build_docker.sh#L58-L60

This is first step in pinning docker images for release


malfet (approved):
LGTM, please include permalink references to places in scripts that supposed to use those

PR #1457, opened 2023-07-12
Support PEP 658/714

thejcannon:
This change is meant to have the PyTorch package index support [PEP 658](https://peps.python.org/pep-0658/) (and [PEP 714](https://peps.python.org/pep-0714/)) by accomplishing two things:

- `index.html` uses both PEP 658/714 tags to denote it supports metadata hosting
  - Currently it uses the value `true`since the S3 bucket doesn't have checksums uploaded (see https://github.com/pytorch/builder/issues/1347 which is specifically about package checksums, but could be extended to metadata checksums)
  - To support PEP 658 and PEP 714, we expose both attributes. Technically PEP 714 supersedes PEP 658, but this is the behavior PyPI has, and is compatible with any package manager who adheres to PEP 658, but not PEP 714.
- The `METADATA` file for any wheel is now uploaded to S3 alongside the wheels in all 3 `upload.sh` scripts (likely needs to be tested thoroughly

Future improvements would be:
- Including checksums for the metadata
- Backporting this change to older wheels (by extracting the METADATA and uploading it to S3)


s3_management/manage.py
```diff
@@ -255,7 +255,13 @@ def to_simple_package_html(
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_","-")))
         for obj in sorted(self.gen_file_list(subdir, package_name)):
-            out.append(f'    <a href="/{obj}">{path.basename(obj).replace("%2B","+")}</a><br/>')
+            attributes = []
+            if obj in self.whls_with_metadata:
+                # Serve the PEP 658 metadata attributes.
+                # For extra juiciness, we should expose the sha256, instead of "true".
+                attributes += 'data-dist-info-metadata="true"'
```
thejcannon, line 262:
```suggestion
                # Serve the PEP 658 and PEP 714 metadata attributes
                attributes += 'data-dist-info-metadata=true'
                attributes += 'data-core-metadata=true'
```

s3_management/manage.py
```diff
@@ -14,7 +14,6 @@
 
 
 S3 = boto3.resource('s3')
-CLIENT = boto3.client('s3')
```
thejcannon, line 17: This was dead code, so drive-by cleanup

matteius: I used that in https://github.com/pytorch/builder/pull/1433 but maybe S3 will have the same effect?

thejcannon: Undoubtedly

thejcannon: Oops, I'm a doofus who can't read :facepalm: 

No `S3` won't be sufficient. I guess we'll just leave it in then :stuck_out_tongue:

thejcannon: Well, looks like it can go again: https://github.com/pytorch/builder/pull/1433#discussion_r1261885338

malfet:
@thejcannon considering that access to the bucket is public, do you mind sharing the output of this script somewhere (I can upload few .metadata files for existing wheels if you want)

thejcannon:
Sorry, for the late response. I'm still very much planning on picking this up, but 2 weeks ago I was on vacation, and then paid the vacation price for it (COVID), so my "return to work" has been slow-going.

thejcannon:
Alright, I'm back on this today. Thanks for everyone's patience.

@malfet I rebased. I probably should've asked as soon as I saw the message for you to upload some metadata files :see_no_evil:

thejcannon:
OK @malfet free free to upload `.metadata` files whenever to kick the tires on this thing.

As far as existing examples go, if you look at https://pypi.org/simple/requests/ and grep for `core-metadata` you'll see 

```html
<a href="https://files.pythonhosted.org/packages/70/8e/0e2d847013cb52cd35b38c009bb167a1a26b2ce6cd6965bf26b47bc0bf44/requests-2.31.0-py3-none-any.whl#sha256=58cd2187c01e70e6e26505bca751777aa9f2ee0b7f4300988b709f44e013003f" data-requires-python=">=3.7" data-dist-info-metadata="sha256=7823e890e9db6f415138badf9744791290ef76e7ec6fd09a3789e8247fffe782" data-core-metadata="sha256=7823e890e9db6f415138badf9744791290ef76e7ec6fd09a3789e8247fffe782">requests-2.31.0-py3-none-any.whl</a>
```

Note that PEP 658 says you can set the attribute to `true` to say "yes I have a metadata file", or set it to the sha256 of the metadata to allow the client to cache even harder. Right now the PR sets `true`, but let me see if we can throw the hash in to be even better.

thejcannon:
Alrighty, everything's been updated since the rebase, and now we set the attribute's value to the sha256 of the metadata file. 

Should be good to test it out. :tada:

vit-zikmund:
Hi @thejcannon, I see the Meta crew has a lot to work on sideways :unamused: 
This could buy us/you some time to throw in this little monkey wrench - [PEP 714](https://peps.python.org/pep-0714/), though. It's very PEP658 related. I actually feel it's so tied to it, it might not even make sense to support just the 568... Look for yourself, it's a real treat :face_with_head_bandage:

thejcannon:
IIRC my change just supports both for maximum compatibility. I'll take a closer look this week

vit-zikmund:
You're right! My bad, sorry.

PR #1468, opened 2023-08-04
Set FORCE_RPATH for PyTorch ROCm wheels

jithunnair-amd:
Using PyTorch ROCm wheels in an environment with ROCm pre-installed can result in errors when running PyTorch. This is because, if LD_LIBRARY_PATH is set in the environment to point to the ROCm installation, it will override the RUNPATH setting for the PyTorch libraries (eg. `libtorch_hip.so`) in the wheel package, thus potentially linking to incompatible libraries.

This PR sets the RPATH instead of RUNPATH, so that LD_LIBRARY_PATH will not interfere with the PyTorch libraries linking to the ROCm libraries packaged in the wheel.

### DIFF
--- a/manywheel/build_rocm.sh
+++ b/manywheel/build_rocm.sh
@@ -14,6 +14,8 @@ export USE_STATIC_NCCL=1
 export ATEN_STATIC_CUDA=1
 export USE_CUDA_STATIC_LINK=1
 export INSTALL_TEST=0 # dont install test binaries into site-packages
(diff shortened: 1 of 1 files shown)


jeffdaily (approved):

pruthvistony (approved):

jithunnair-amd:
@atalman Can you please review and approve? This will resolve a usability issue with Pytorch nightly wheels for ROCm.

atalman (approved):
LGTM

PR #1470, opened 2023-08-08
Decouple aarch64 ci setup and build steps

atalman:
We want to reuse setup script for building aarch64 torchvision, torchaudio and torchtext. Hence decoupling setup from pytorch build step.


xncqr (approved):
I'm working on the aarch64 docker image that will have everything needed for aarch64 manywheel and conda, removing the need for aarch64_ci_setup.sh in the end. PR to be in shortly for that.

To move forward while that PR is in work, this is a good interim.

PR #1522, opened 2023-09-05
Torchmetrics in S3 Index

osalpekar:
We will need the stable torchmetrics wheel in the S3 index, since torchrec depends on it. This is similar to how pytorch depends on numpy, etc. and these binaries need to be hosted in our index when uses try to pip install from download.pytorch.org.

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -85,6 +85,7 @@ PACKAGE_ALLOW_LIST = {
     "torchcsprng",
     "torchdata",
     "torchdistx",
+    "torchmetrics",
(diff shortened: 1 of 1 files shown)


osalpekar:
cc @henrylhtsang @colin2328

henrylhtsang:
Note that we only need torchmetrics=1.0.3. The newest torchmetrics breaks some tests.

seemethere (approved):

osalpekar:
Added torchmetrics wheel to our index, merging this PR

PR #1526, opened 2023-09-08
update NCCL==2.18.3

ptrblck:
Move NCCL to `2.18.3` as it was used in `pytorch:main` before the most recent update to `2.18.5`.
Note that NCCL did not release `2.18.5` binaries yet, so we should stick to the latest `2.18.3` version in `torch==2.1.0`.

CC @atalman

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -32,8 +32,6 @@ else
 
     if [[ ${TARGET_OS} == 'windows' ]]; then
         python  ./test/smoke_test/smoke_test.py ${TEST_SUFFIX}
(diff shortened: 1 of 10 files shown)


ptrblck:
@malfet @atalman 

`pytorch/pytorch` updated NCCL in https://github.com/pytorch/pytorch/pull/107883 which also landed in the `2.1.0` release candidates. E.g.: https://github.com/pytorch/pytorch/tree/v2.1.0-rc3/third_party

Since the `2.18.5` binaries are not released yet we would diverge in `torch==2.1.0` between the binaries and a source build.
Should we revert https://github.com/pytorch/pytorch/pull/107883 in `v2.1.0-rc3`?

ptrblck:
Closing in favor of: https://github.com/pytorch/builder/pull/1527

PR #1527, opened 2023-09-08
update NCCL==2.18.3

ptrblck:
https://github.com/pytorch/builder/pull/1526 with the right base branch.

CC @atalman


ptrblck:
Discussed it with @atalman in our sync again and we won't align the binaries to source builds in this release, so closing it.

PR #1528, opened 2023-09-11
Avoid PRE_CXX11_SYMBOLS false positives

cyyever:
As pytorch moving to C++17, the binary contains both "std::basic_string_view" and "std::__cxx11::basic_string<", change the pattern to avoid finding out std::basic_string_view, causing false positives.

cc @atalman @osalpekar @malfet @huydhn

### DIFF
--- a/check_binary.sh
+++ b/check_binary.sh
@@ -116,7 +116,7 @@ if [[ "$(uname)" != 'Darwin' ]]; then
   #
   # To check whether it is using cxx11 ABI, check non-existence of symbol:
   PRE_CXX11_SYMBOLS=(
-    "std::basic_string"
+    "std::basic_string<"
     "std::list"
   )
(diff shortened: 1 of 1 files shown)


cyyever:
@huydhn Can help review？

huydhn (approved):

malfet (changes requested):
Hmm, move to C++17 should not in any way affect what symbols we expect to see, should it?

cyyever:
@malfet Unfortunately, it uses substring match to find PRE_CXX11_SYMBOLS, however, std::string_view was falsely recognised as PRE_CXX11_SYMBOLS, preventing adoption of std::string_view.

malfet (approved):

malfet:
@cyyever thank you for the explanation, merged.

PR #1531, opened 2023-09-12
S3 Management Job Outside Docker

osalpekar:
To update the S3 index, we build a docker image and run that from a GHA in `pytorch/pytorch`. This docker image has not been rebuilt in 3 months, and so changes to `s3_management` are not reflected. The docker image also basically just installed requirements and run this python script, so rebuilding and maintaining that image just added overhead. We will do the S3 management from this new GHA that runs the script directly instead of using a docker image.

### DIFF
--- a/.github/workflows/update-s3-html.yml
+++ b/.github/workflows/update-s3-html.yml
@@ -1,7 +1,6 @@
 name: Update S3 HTML indices for download.pytorch.org
 
 on:
-  pull_request:
   schedule:
     # Update the indices every 30 minutes
     - cron: "*/30 * * * *"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-s3-html.yml
```diff
@@ -0,0 +1,36 @@
+name: Update S3 HTML indices for download.pytorch.org
+
+on:
+  pull_request:
```
osalpekar, line 4: Will remove this prior to merging.

huydhn (approved):

malfet:
@osalpekar please use `Squash and merge` rather than `merge` and  commit description before landing it as https://github.com/pytorch/builder/commit/9f365204ffce6d37c5447e07f8a5db291120ec16 looks as follows:
```
* S3 Management Job Outside Docker

* job name

* remove failfast

* no matrix

* inherit secrets

* spacing?

* random nits

* add back secrets

* add back matrix

* export env vars correctlty

* Update update-s3-html.yml
```

PR #1541, opened 2023-09-17
Update builder images to ROCm5.7

jithunnair-amd:
cc @jeffdaily @BLOrange-AMD

PR #1543, opened 2023-09-19
Update docker build convenience scripts to ROCm5.7

jithunnair-amd:
Companion to https://github.com/pytorch/builder/pull/1541


jithunnair-amd:
These scripts are not called via any workflows etc., they're just convenience scripts.

PR #1370, opened 2023-03-29
[aarch64] acl build script updates for multi isa build with armv8-a base architecture

snadampal:
This PR fixes the issue https://github.com/pytorch/pytorch/issues/97226

PR #1552, opened 2023-09-29
[aarch64] Pin ACl to v22.11

atalman:
This switches ACL back to working configuration.

At a time of this PR:
https://github.com/pytorch/builder/pull/1370

This code was added in Compute Library v23.05 :
```
if 'v8a' in env['arch']:
        print("INFO: multi_isa armv8-a architecture build doesn't enable __ARM_FEATURE_FP16_VECTOR_ARITHMETIC. Use armv8.2-a or beyond to enable FP16 vector arithmetic support")
        env.Append(CXXFLAGS = ['-march=armv8-a']) # note: this will disable fp16 extension __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
```
Ref:
https://github.com/ARM-software/ComputeLibrary/commit/6c713f090601ea839d944a30888ea56eb2f43988#diff-d78c5c377c03dd251edc72cdb0ae7d9986fa08361eb392187218f8c670c0bc4eR311

### DIFF
--- a/aarch64_linux/aarch64_wheel_ci_build.py
+++ b/aarch64_linux/aarch64_wheel_ci_build.py
@@ -20,7 +20,10 @@ def build_ArmComputeLibrary(git_clone_flags: str = "") -> None:
     '''
     print('Building Arm Compute Library')
     os.system("cd / && mkdir /acl")
-    os.system(f"git clone https://github.com/ARM-software/ComputeLibrary.git -b v23.05.1 {git_clone_flags}")
+    os.system(f"git clone https://github.com/ARM-software/ComputeLibrary.git -b v22.11 {git_clone_flags}")
+    os.system('sed -i -e \'s/"armv8.2-a"/"armv8-a"/g\' ComputeLibrary/SConscript; '
+              'sed -i -e \'s/-march=armv8.2-a+fp16/-march=armv8-a/g\' ComputeLibrary/SConstruct; '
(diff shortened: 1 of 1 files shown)


malfet (commented):
Sure, those I've tried this one few weeks back and it did not work

This is why I've collected the backtrace in https://github.com/pytorch/pytorch/issues/109312 to show that it comes from https://github.com/oneapi-src/oneDNN/blob/99c3fb44488922554696276dc88e8b6279eb662d/src/common/utils.cpp#L113

snadampal:
@malfet , it's crashing even without ACL. I have built the wheels with `USE_MKLDNN=ON` but `USE_MKLDNN_ACL=OFF`, and observed the crash. So, i think it's the onednn build we need to check.

snadampal:
@malfet , please don't revert the ACL version. Please check my comment on the original issue
https://github.com/pytorch/pytorch/issues/109312#issuecomment-1741362365

malfet:
> @malfet , it's crashing even without ACL. I have built the wheels with `USE_MKLDNN=ON` but `USE_MKLDNN_ACL=OFF`, and observed the crash. So, i think it's the onednn build we need to check.

@snadampal sorry, I should have made it clear in the issue, that this is not an ACL problem, but rather a oneDNN one

malfet:
No longer needed

PR #1561, opened 2023-10-02
[BE] Delete unused batch script

malfet:
Those scripts does not seem to be invoked at all


malfet:
Actually, they are still used

PR #1572, opened 2023-10-13
Add ROCM_PATH env var to Dockerfile for ROCm5.7 issue with finding HIP

jithunnair-amd:
More info in commit description [here](https://github.com/ROCmSoftwarePlatform/builder/commit/66369a64c38ef5db6d2cd0fc7701688f872855f1).

This workaround will be needed once pytorch/pytorch LoadHIP.cmake is updated for ROCm6.0 changes.
This workaround is not needed for ROCm6.0 and above, as HIP cmake logic was fixed for those versions, and should be removed once ROCm5.7 is not in the support matrix for wheel builds anymore.

cc @pruthvistony


jithunnair-amd:
CUDA docker build CI failure is unrelated. Merging.

pruthvistony (commented):
Temp workaround.

PR #1576, opened 2023-11-02
Upload to docker.io when pushing release tag

huydhn:
I notice that there is no `-2.1` tag for conda and libtorch images on docker hub:
* https://registry.hub.docker.com/r/pytorch/conda-builder/tags
* https://registry.hub.docker.com/r/pytorch/libtorch-cxx11-builder/tags

On the other hand, manywheel images have it:
* https://registry.hub.docker.com/r/pytorch/manylinux-builder/tags

It turns out that the images are pushed only when some file are touched on the release branch, not when RC is cut.  Thus, this change added the same logic as docker-release workflow from PyTorch https://github.com/pytorch/pytorch/blob/main/.github/workflows/docker-release.yml#L29C3-L29C196 to address this issue.

### DIFF
--- a/.github/workflows/build-conda-images.yml
+++ b/.github/workflows/build-conda-images.yml
@@ -24,7 +24,7 @@ env:
   DOCKER_BUILDKIT: 1
   DOCKER_ID: ${{ secrets.DOCKER_ID }}
   DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
-  WITH_PUSH: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release')) }}
+  WITH_PUSH: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release') || (startsWith(github.event.ref, 'refs/tags/')) }}
 
 jobs:
   build-docker:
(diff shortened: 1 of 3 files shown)


malfet (changes requested):
I don't think we ever pushed a release tag in docker repo. And even if we do, why should we kick another build, as it shouldn't produce a binary any different than the one from a previous build

huydhn:
> I don't think we ever pushed a release tag in docker repo. And even if we do, why should we kick another build, as it shouldn't produce a binary any different than the one from a previous build

Is pushing the release tag the step we need to do to have different Docker images for different release branch cut?  Without this, there might be Docker images with release tags, for example, you won't find `-2.1` in:

* https://registry.hub.docker.com/r/pytorch/conda-builder/tags
* https://registry.hub.docker.com/r/pytorch/libtorch-cxx11-builder/tags

It's the same binary Docker image indeed, but has a different tags I think.

malfet:
> Is pushing the release tag the step we need to do to have different Docker images for different release branch cut?

No, as we already have https://registry.hub.docker.com/r/pytorch/manylinux-builder/tags?page=1&name=cuda11.8-2.1

The reason why we rely on `-rc1` tags for `pytorch/pytorch` repo, is that we don't want to kick full binary builds on every commit to the branch, but for builder we are doing it already, so we can use tags as some sort of milestones for ourselves, but we do not need to kick a new build, as it is already there

huydhn:
Closing this, as this is not needed anymore after chatting with @malfet 

* We will implement workflow dispatch mechanism instead
* If 2.1 tag is needed, it can be pushed manually pointing to the existing images

PR #1581, opened 2023-11-12
fix: typo

0x804d8000:


0x804d8000:
Hi, @ptrblck @albanD maybe you can take a look at this PR? Thanks in advance.

kit1980:
I tried shellcheck on the file, and it finds the issue:
> SC2153: Possible misspelling: PREIX may not be assigned, but PREFIX is.

We probably should adopt shellcheck linter for the repo.

malfet (approved):

PR #1586, opened 2023-11-14
Changes needed for core enablement of 3.12 binaries

albanD:
Namely for https://github.com/pytorch/pytorch/pull/112882 in core.


aarch64_linux/aarch64_ci_setup.sh
```diff
@@ -19,7 +19,16 @@ curl -L -o /mambaforge.sh https://github.com/conda-forge/miniforge/releases/late
...
+conda config --set ssl_verify False
+conda create -y -c conda-forge -n aarch64_env python=${DESIRED_PYTHON}
+conda activate aarch64_env
+conda install -y -c conda-forge numpy
+conda install -y -c conda-forge pyyaml
+conda install -y -c conda-forge patchelf
+conda install -y -c conda-forge pygit2
+conda install -y -c conda-forge openblas
+conda install -y -c conda-forge ninja
+conda install -y -c conda-forge scons
```
malfet, line 32: do you mind merging it into one command and pinning versions here?
```suggestion
conda install -y -c conda-forge numpy==1.26 pyyaml patchelf pygit2 openblas ninja scons
```

malfet (approved):

atalman (approved):

aarch64_linux/aarch64_ci_setup.sh
```diff
@@ -19,7 +19,16 @@ curl -L -o /mambaforge.sh https://github.com/conda-forge/miniforge/releases/late
 chmod +x /mambaforge.sh
 /mambaforge.sh -b -p /opt/conda
 rm /mambaforge.sh
-/opt/conda/bin/conda config --set ssl_verify False
-/opt/conda/bin/conda install -y -c conda-forge python=${DESIRED_PYTHON} numpy pyyaml setuptools patchelf pygit2 openblas ninja scons
+source /opt/conda/etc/profile.d/conda.sh
```
janeyx99, line 22: Can you explain why this is necessary/what this changes for the sake of understanding?

aarch64_linux/aarch64_ci_setup.sh
```diff
@@ -19,7 +19,16 @@ curl -L -o /mambaforge.sh https://github.com/conda-forge/miniforge/releases/late
...
+conda config --set ssl_verify False
+conda create -y -c conda-forge -n aarch64_env python=${DESIRED_PYTHON}
```
janeyx99, line 24: lol this is also an important distinction that an env is now created--not sure if this will have other implications

janeyx99: (to be clear, this seems more correct)

aarch64_linux/aarch64_ci_setup.sh
albanD: I did that because it makes it a lot simpler to understand which package conda cannot find and leads to 200lines of "un-resolvable unvironment". happy to change it back if you prefer.

aarch64_linux/aarch64_ci_setup.sh
albanD: This is needed to be able to create the env below. Answering there.

aarch64_linux/aarch64_ci_setup.sh
albanD: The current code is inplace modifying the env from being python 3.10 to another python version as asked. For some reason that worked for 3.8/3.9/3.11 but 3.12 changed which packages are in the default distribution and so just doing `/opt/conda/bin/conda install -y -c conda-forge python=${DESIRED_PYTHON}` fails.

aarch64_linux/aarch64_ci_setup.sh
albanD: Updated and added all the pins based on the last run from the core CI

PR #1611, opened 2023-11-16
Release validations using release version matrix

atalman:
Allows you to specify release version to validate. This is first step in supporting multiple release versions at once.
If version is not specified. The validations will only validate torch against version coming from matrix. This is same as current behaviour. 
IIf version is specified. System will validate with version provided.

Test with version 2.1.1:
https://github.com/pytorch/builder/actions/runs/6895835474

Test with no version specified:
https://github.com/pytorch/builder/actions/runs/6896242630/job/18761933361


huydhn (approved):
LGTM!

PR #1622, opened 2023-11-21
Fix poetry and pypi validations when version is specified

atalman:
1. restrict poetry validation to only release, add support for version. Please note testing using test or nightly channel does not makes much sence. Hence keeping it to release only
2. pypi validation should allow version to be specified
3. fix typo for macos validations

Test: https://github.com/pytorch/builder/actions/runs/6945113935


huydhn (approved):
LGTM!

PR #1627, opened 2023-11-28
[aarch64] ACL, OpenBLAS and mkldnn updates for PyTorch 2.2

snadampal:
Note# ~~This PR has a dependency on updating the oneDNN version to v3.3 (via ideep submodule to v3.3)~~
ideep submodule update is done, so, this PR can be merged anytime now.

This PR is for:
ACL - build with fixed format kernels 
OpenBLAS - upgrade the version to 0.3.25
numpy - upgrade version to 1.26.2
and mkldnn - cleanup the patches that are already upstreamed.


malfet:
@snadampal so you are updaing OpenBLAS when build from source, but not when installed from conda forge? Why is that?

malfet (approved):

PR #1628, opened 2023-12-01
Add aarch64 conda env lib to LD_LIBRARY_PATH

huydhn:
After the change on https://github.com/pytorch/builder/pull/1586, nightly aarch64 wheel fails to find `libopenblas.so` which is now installed under `/opt/conda/envs/aarch64_env/lib/` instead of the base conda `/opt/conda/lib`.  Using CPU nightly wheels on aarch64 from Nov 16 then ends up with the error as described in https://github.com/pytorch/pytorch/issues/114862: `Calling torch.geqrf on a CPU tensor requires compiling PyTorch with LAPACK. Please use PyTorch built with LAPACK support`.  The error can be found on night build log https://github.com/pytorch/pytorch/actions/runs/6887666324/job/18735230109#step:15:4933

Fixes https://github.com/pytorch/pytorch/issues/114862

### DIFF
--- a/aarch64_linux/aarch64_ci_setup.sh
+++ b/aarch64_linux/aarch64_ci_setup.sh
@@ -6,8 +6,9 @@ set -eux -o pipefail
 
 CONDA_PYTHON_EXE=/opt/conda/bin/python
 CONDA_EXE=/opt/conda/bin/conda
+CONDA_ENV_NAME=aarch64_env
 PATH=/opt/conda/bin:$PATH
-LD_LIBRARY_PATH=/opt/conda/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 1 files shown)


huydhn:
> Hmm, we should not be installing OpenBLAS from Anaconda, but rather build it form source

We are using openblas from conda atm.  Building it from source is feasible, and needs to be done via PyTorch third_party deps I guess.  Let me take a look at that.

In the meantime, should we land this change to fix nightly?

malfet:
@snadampal did we do any perf benchmark between one from anaconda and one that was build from source?
Small challenge I have with depending on condaforge, is that I don't know what feature were included and not included into the build. For example, was it build with pthreads or openmp support for mulithreading?

snadampal:
@malfet , thank you. yeah, we also had to build the openBLAS package with enough threads (static config) to make sure it scales beyond the host machine config.  torch 2.1 wheel performance was on par with torch 2.0.1 in our testing, but we will check conda package perf explicitly.

PR #1630, opened 2023-12-02
[aarch64] cleanup mkldnn patching

snadampal:
pytorch is moved to oneDNN v3.3.2 and the old patches are not applicable any more, hence removing them.

Note# However, I haven't tested whether oneDNN v3.3.2 fixes the issue https://github.com/pytorch/pytorch/issues/111695


malfet (changes requested):
Sorry, is it possible to have a test plan of sorts?

Also, how does it work right now, see https://github.com/pytorch/pytorch/actions/runs/7149862957/job/19472525415 for example?

malfet (approved):
Oh, I see, we use `os.system`, so error codes are not propagated back, LGTM, but will rework all `os.system` to `subprocess.check_call` to surface the error

PR #1631, opened 2023-12-03
[aarch64] patch mkldnn acl inner product to accelerate torch.compile() for bert

snadampal:


malfet (changes requested):
Sorry, patch mkldnn to fix build errors is probably appropriate, but feature work should be done on the appropriate repository.

So lets work with onednn to get this fix landed into their default branch and then probably update a release branch


Also, can you please clarify, if this is a regression from previos release or just a new feature?

snadampal:
Hi @malfet , this is the [PR](https://github.com/oneapi-src/oneDNN/pull/1768) for oneDNN repo which will be available in oneDNN 3.4. But the oneDNN3.4 may not be available for PyTorch 2.2 timeline. and this change is mandatory to get torch.compile() performance improvements on aarch64.

malfet:
@snadampal I understand the sentiment, but:
 - oneDNN PR is not merged yet
 - It's already past the deadline for 2.2 branch cut (and this is a feature work isn't it?)
 - There are no test plan (and CI does not test it at all)
 - Builder is not the right place to apply patches, as it prevents users, who will checkout 2.2 release branch from build package identical to the one being published (all previous patches were to unblock the builds/fix fundamental regressions)

PR #1645, opened 2023-12-15
Fix validation workflow on aarch64 with conda 23.11.0 and GLIBC_2.25

huydhn:
The latest conda 23.11.0 pulls in some dependencies of conda-libmamba-solver that require GLIBC_2.25, which is not available in the current aarch64 image causing the subsequence git command to fail. Basically, they don't work with CentOS 7 which AML 2 is based on https://github.com/ContinuumIO/anaconda-issues/issues/12822

The simple fix is to not set LD_LIBRARY_PATH which picks up libraries in /opt/conda/lib before the smoke test.  This step is not needed at this point.

### Testing

https://github.com/pytorch/builder/actions/runs/7227335815/job/19694776867

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -39,11 +39,14 @@ else
         ${PWD}/check_binary.sh
     fi
 
-    # NB: The latest conda 23.11.0 pulls in some dependencies of conda-libmamba-solver that
-    # require GLIBC_2.25, which is not available in the current aarch64 image causing the
-    # suqsequence git command to fail. Basically, they don't work with CentOS 7 which AML 2
-    # is based on https://github.com/ContinuumIO/anaconda-issues/issues/12822
-    unset LD_LIBRARY_PATH
+    if [[ ${TARGET_OS} == 'linux-aarch64' ]]; then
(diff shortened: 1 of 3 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -39,6 +39,12 @@ else
         ${PWD}/check_binary.sh
     fi
 
+    # NB: The latest conda 23.11.0 pulls in some dependencies of conda-libmamba-solver that
+    # require GLIBC_2.25, which is not available in the current aarch64 image causing the
+    # suqsequence git command to fail. Basically, they don't work with CentOS 7 which AML 2
+    # is based on https://github.com/ContinuumIO/anaconda-issues/issues/12822
+    unset LD_LIBRARY_PATH
+
```
atalman, line 47: Maybe better approach here would be to 
- Save path
- Unset
- Set it back before finishing

atalman: Also please add check for
https://github.com/pytorch/builder/blob/main/.github/workflows/validate-aarch64-linux-binaries.yml#L92

atalman (approved):

PR #1647, opened 2023-12-19
Update builder images to ROCm6.0

jithunnair-amd:
Also add gfx942 (MI300X) to list of archs to build for.


jithunnair-amd:
@jeffdaily please review to make sure we're not missing any ROCm6.0-specific updates. The builder images built fine.

jeffdaily (approved):

PR #1664, opened 2024-01-05
Fix quotation issues when migrating from python file to one line format

nWEIdia:
Sorry, looks like the last line had an issue while porting it from multi-line python file to one-line. 

Side question: when does this file get used? Is it only used during release binary generation/testing?


nWEIdia:
cc @atalman @malfet @huydhn @eqy @ptrblck

eqy (approved):

malfet (approved):

malfet:
@nWEIdia yes, only during binaries testing, so we would not see it until next nightly

nWEIdia:
> @nWEIdia yes, only during binaries testing, so we would not see it until next nightly

Thank you @malfet! I will also keep an eye on it in case anything else falls apart.

PR #1670, opened 2024-01-08
Build nccl 2.19.3 for CUDA 11.8 and 12.1

atalman:
Fix: https://github.com/pytorch/pytorch/issues/116977

Nccl 2.19.3 don't exist for cuda 11.8 and cuda 12.1. Refer to https://docs.nvidia.com/deeplearning/nccl/release-notes/rel_2-19-3.html#rel_2-19-3 CUDA 12.0, 12.2, 12.3 are supported.

Hence we do manual build. Follow this build process:
https://github.com/NVIDIA/nccl/tree/v2.19.3-1?tab=readme-ov-file#build

We want nccl version be exactly the same as installed here:
https://github.com/pytorch/pytorch/blob/main/.github/scripts/generate_binary_build_matrix.py#L45

### DIFF
--- a/common/install_cuda.sh
+++ b/common/install_cuda.sh
@@ -33,13 +33,13 @@ function install_118 {
     rm -rf tmp_cudnn
 
     # NCCL license: https://docs.nvidia.com/deeplearning/nccl/#licenses
-    mkdir tmp_nccl && cd tmp_nccl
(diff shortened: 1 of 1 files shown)


huydhn (approved):

malfet (approved):
Looks good, but please refactor repeated code into its own function

PR #1696, opened 2024-02-03
ci: aarch64 linux: fix torch performance regression with conda openblas package

snadampal:
changing the conda openblas package from pthread version to openmp version to match torch openmp runtime.  The pthread version was conflicting with the openmp runtime and causing thread over-subscription and performance degradation.

This fixes the regression introduced when the wheel builder was moved from local openblas builds to conda package in PyTorch 2.1.


malfet (approved):

malfet:
LGTM, but I wonder if we can test for it somehow

snadampal:
I have created this [issue](https://github.com/pytorch/pytorch/issues/119374) with the test details and tagged here.

PR #1697, opened 2024-02-06
Remove custom PyTorch build dependency logic on 3.11

huydhn:
Fixes https://github.com/pytorch/pytorch/issues/114721, we had an obsolete custom logic for 3.11 when it comes to install build dependencies.  Specifically, the 3.11 if branch  doesn't include `nomkl`.

I think this can be removed because the else branch works for 3.12.

If this works out, I will cherry pick the change to 2.2.1.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -142,7 +142,7 @@ if [[ "$cuda_ver" != 'cpu' ]]; then
     fi
 fi
 
-# Check that OpenBlas is not linked to on Macs
+# Check that OpenBlas is not linked to on MacOS
 if [[ "$(uname)" == 'Darwin' ]]; then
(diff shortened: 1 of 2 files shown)


huydhn:
Testing on PyTorch https://github.com/pytorch/pytorch/pull/119246

atalman (approved):

huydhn:
Confirmed at the wheel torch-2.3.0.dev20240206-cp311-none-macosx_11_0_arm64.whl from https://github.com/pytorch/pytorch/pull/119246 is working

```
python -c "import os;import torch;print(torch.backends.openmp.is_available())"
True
```

I will merge and cherry pick this one.

PR #1698, opened 2024-02-06
Pin conda-build to 3.28.4

huydhn:
Conda build 24.1.0 release is botched with a pretty serious bug https://github.com/conda/conda-build/issues/5167 where successful builds are reported as failures.  This causes nightly conda build to fail on MacOS.  It's a bit weird that Linux and Windows conda builds are still working.

There might be a better workaround coming from https://github.com/conda/conda-build/issues/5167, but it's easy to pin the version for the time being.

### DIFF
--- a/conda/build_pytorch.sh
+++ b/conda/build_pytorch.sh
@@ -208,7 +208,8 @@ if [[ "$(uname)" == 'Darwin' ]]; then
         "$miniconda_sh" -b -p "$tmp_conda" && \
         rm "$miniconda_sh"
     export PATH="$tmp_conda/bin:$PATH"
-    retry conda install -yq conda-build
+    # TODO(huydhn): We can revert the pin after https://github.com/conda/conda-build/issues/5167 is resolved
+    retry conda install -yq conda-build=3.28.4
 elif [[ "$OSTYPE" == "msys" ]]; then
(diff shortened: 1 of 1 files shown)


huydhn:
Testing in https://github.com/pytorch/pytorch/pull/119252

atalman (approved):

conda/build_pytorch.sh
```diff
@@ -219,7 +220,8 @@ elif [[ "$OSTYPE" == "msys" ]]; then
     pushd $tmp_conda
     export PATH="$(pwd):$(pwd)/Library/usr/bin:$(pwd)/Library/bin:$(pwd)/Scripts:$(pwd)/bin:$PATH"
     popd
-    retry conda install -yq conda-build
+    # TODO(huydhn): We can revert the pin after https://github.com/conda/conda-build/issues/5167 is resolved
```
atalman, line 223: Since failure is with MacOS only we probably don't need to apply this constraint to windows. Only MacOS should be enought

malfet (approved):

conda/build_pytorch.sh
huydhn: I hope that conda doesn't put us into a situation where both pinning an older version and using the latest version fail :)

PR #1717, opened 2024-03-06
aarch64: apply the cherrypicked onednn PR-1768

snadampal:
This is to improve the torch.compile() perf by 5.8x on AWS Graviton3 instances. This patching is required till PyTorch oneDNN is upgraded to v3.4.

### DIFF
--- a/aarch64_linux/aarch64_wheel_ci_build.py
+++ b/aarch64_linux/aarch64_wheel_ci_build.py
@@ -110,6 +110,9 @@ if __name__ == '__main__':
     print("Applying mkl-dnn patch to fix crash due to /sys not accesible")
     os.system("cd /pytorch/third_party/ideep/mkl-dnn && patch -p1 < /builder/mkldnn_fix/fix-xbyak-failure.patch")
 
+    print("Applying mkl-dnn patch to improve torch.compile() perf")
+    os.system("cd /pytorch/third_party/ideep/mkl-dnn && patch -p1 < /builder/mkldnn_fix/onednn-pr1768-aarch64-add-acl-sbgemm-inner-product-primitive.patch")  # noqa: E501
+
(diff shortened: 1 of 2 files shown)


atalman (approved):

malfet:
Why have we landed this change? If it missed the deadline for oneDNN branch cut, then it should wait until the next release or needs to be cherry-picked into one

aarch64_linux/aarch64_wheel_ci_build.py
```diff
@@ -110,6 +110,9 @@ def parse_arguments():
     print("Applying mkl-dnn patch to fix crash due to /sys not accesible")
     os.system("cd /pytorch/third_party/ideep/mkl-dnn && patch -p1 < /builder/mkldnn_fix/fix-xbyak-failure.patch")
 
+    print("Applying mkl-dnn patch to improve torch.compile() perf")
+    os.system("cd /pytorch/third_party/ideep/mkl-dnn && patch -p1 < /builder/mkldnn_fix/onednn-pr1768-aarch64-add-acl-sbgemm-inner-product-primitive.patch")  # noqa: E501
```
malfet, line 114: Not to author, reviewer: I don't know why those were accepted initially, but one should avoid using `os.system` as much as possible, as it gives no signal back on whether the underlying command succeeded or failed, this way making regression detection a much harder process. (i.e. one can add `os.system("fokjhdsfkjhsef");` anywhere in this script and it will be just a no-op)

`subprocess.check_call` is a better way of doing that, for example like that:
https://github.com/pytorch/builder/blob/ab5fc904118395902a29ba74df20ca2a5e937694/aarch64_linux/aarch64_wheel_ci_build.py#L111-L112

This way typo in patch name or failure to apply the patch will result in the failure

atalman: Sounds good we should probably change it in both palces line 111 and 114

PR #1733, opened 2024-03-13
chore: Remove LLVM build; was used by NNC which is dead and has since been removed from containers

bryantbiggs:
Remove LLVM build; was used by NNC which is dead and has since been removed from containers

- Removed from containers in https://github.com/pytorch/builder/pull/1445
- Relates https://github.com/pytorch/pytorch/issues/105092


bryantbiggs:
@atalman / @malfet is this a valid change or no?

PR #1737, opened 2024-03-14
Add builder for s390x

AlekseiNikiforovIBM:
It's based on aarch64, but bison and yasm are missing


huydhn:
Is there even a `linux.s390x` runner on PyTorch atm? I couldn’t find it

huydhn (approved):

PR #1747, opened 2024-03-20
Update nightly wheel build numpy to 2.0

atalman:
Test In: https://github.com/pytorch/pytorch/pull/121979

### DIFF
--- a/windows/internal/smoke_test.bat
+++ b/windows/internal/smoke_test.bat
@@ -52,7 +52,13 @@ if errorlevel 1 exit /b 1
 
 set "PATH=%CD%\Python%PYTHON_VERSION%\Scripts;%CD%\Python;%PATH%"
 
-pip install -q numpy protobuf
+
+if "%DESIRED_PYTHON%" == "3.12" pip install -q --pre numpy==2.0.0b1 protobuf
+if "%DESIRED_PYTHON%" == "3.11" pip install -q --pre numpy==2.0.0b1 protobuf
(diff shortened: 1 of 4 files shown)


windows/internal/smoke_test.bat
```diff
@@ -52,7 +52,13 @@ if errorlevel 1 exit /b 1
 
 set "PATH=%CD%\Python%PYTHON_VERSION%\Scripts;%CD%\Python;%PATH%"
 
-pip install -q numpy protobuf
+
+if "%DESIRED_PYTHON%" == "3.12" pip install -q --pre numpy==2.0.0b1 protobuf
+if "%DESIRED_PYTHON%" == "3.11" pip install -q --pre numpy==2.0.0b1 protobuf
+if "%DESIRED_PYTHON%" == "3.10" pip install -q --pre numpy==2.0.0b1 protobuf
+if "%DESIRED_PYTHON%" == "3.9" pip install -q --pre numpy==2.0.0b1 protobuf
```
albanD, line 59: nit: maybe we can make this an else to avoid having the same lines multiple times?

albanD (approved):
Sound good!

windows/internal/smoke_test.bat
atalman: Unfortunately this is windows limitation, see: https://superuser.com/questions/1641210/multiple-if-else-conditions-in-batch-file

PR #1758, opened 2024-03-28
chore: Drop kepler PTX generation since its a deprecated arch

bryantbiggs:
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#gpu-feature-list


bryantbiggs:
@ptrblck what are your thoughts on dropping this?

PR #1779, opened 2024-04-08
[Release only] Pin PYTORCH_BUILD_NUMBER for final RC

atalman:
Previous builds all used : PYTORCH_BUILD_NUMBER = 0
You can see this from index.json of conda package:
```
  "arch": "x86_64",
  "build": "py3.9_cuda12.1_cudnn8.9.2_0",
  "build_number": 0,
```


atalman:
Will follow up with Anaconda team on this. i think we perhaps require to have Build = 0

PR #1799, opened 2024-04-24
Update promote.sh

spcyppt:
Update FBGEMM version to be promoted.

### DIFF
--- a/release/promote.sh
+++ b/release/promote.sh
@@ -18,7 +18,7 @@ TENSORRT_VERSION=${TENSORRT_VERSION:-2.2.0}
 # be without rc suffix, fbgemm_gpu-0.6.0+cpu-cp311-cp311, and that's the one
 # ready to be promoted. So, keeping a + here in the version name allows the
 # promote script to find the correct binaries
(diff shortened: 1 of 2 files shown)


release/promote.sh
```diff
@@ -18,7 +18,7 @@ TENSORRT_VERSION=${TENSORRT_VERSION:-2.2.0}
 # be without rc suffix, fbgemm_gpu-0.6.0+cpu-cp311-cp311, and that's the one
 # ready to be promoted. So, keeping a + here in the version name allows the
 # promote script to find the correct binaries
-FBGEMMGPU_VERSION=${FBGEMMGPU_VERSION:-0.6.0+}
+FBGEMMGPU_VERSION=${FBGEMMGPU_VERSION:-0.7.0+}
```
spcyppt, line 21: Note: Only our rc0 still has -rc0 suffix. We have changed the name convention and removed suffix from rc1 onwards.

huydhn:
You would need to update `release/release_versions.sh` too I think, similar to https://github.com/pytorch/builder/pull/1793/files#diff-095dd12064d9409272ab08fe63c76935264c2070b0055fc8d961807bcff90078

huydhn (approved):

PR #1802, opened 2024-04-25
Update s390x builder

AlekseiNikiforovIBM:
Update binary checks for s390x build

### DIFF
--- a/.github/workflows/build-manywheel-images.yml
+++ b/.github/workflows/build-manywheel-images.yml
@@ -137,18 +137,3 @@ jobs:
       - name: Build Docker Image
         run: |
           manywheel/build_docker.sh
-  build-docker-cpu-s390x:
-    runs-on: linux.s390x
-    env:
-      GPU_ARCH_TYPE: cpu-s390x
-    steps:
-      - name: Checkout PyTorch
-        uses: actions/checkout@v3
-      - name: Authenticate if WITH_PUSH
-        run: |
-          if [[ "${WITH_PUSH}" == true ]]; then
-            echo "${DOCKER_TOKEN}" | docker login -u "${DOCKER_ID}" --password-stdin
-          fi
-      - name: Build Docker Image
-        run: |
-          manywheel/build_docker.sh
--- a/manywheel/Dockerfile_s390x
+++ b/manywheel/Dockerfile_s390x
@@ -1,7 +1,8 @@
 FROM --platform=linux/s390x docker.io/redhat/ubi9 as base
 
-# earliest available version in ubi9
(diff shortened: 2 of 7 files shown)


.github/workflows/build-manywheel-images.yml
```diff
@@ -137,18 +137,3 @@ jobs:
       - name: Build Docker Image
         run: |
           manywheel/build_docker.sh
-  build-docker-cpu-s390x:
```
huydhn, line 140: Is this a correct understanding that we can now remove this and let Docker pull figure out the correct CPU arch (s390x) corresponding to this change https://github.com/pytorch/pytorch/pull/120398/commits/2928b2c4d1bfa0ac03881b87cb965432ac381da2?

manywheel/Dockerfile_s390x
```diff
@@ -1,18 +1,15 @@
-FROM --platform=linux/s390x docker.io/redhat/ubi9 as base
+FROM --platform=linux/s390x docker.io/ubuntu:24.04 as base
```
huydhn, line 1: Note:  If it works, we can stick with ubuntu for now.  Otherwise, let's use `AlmaLinux 8` as mentioned in https://github.com/pypa/manylinux per our discussion.

manywheel/build.sh
```diff
@@ -15,7 +15,7 @@ case "${GPU_ARCH_TYPE:-BLANK}" in
     rocm)
         bash "${SCRIPTPATH}/build_rocm.sh"
         ;;
-    cpu | cpu-cxx11-abi)
+    cpu | cpu-cxx11-abi | cpu-s390x)
```
huydhn, line 18: Should `cpu-s390x` be removed from here too?

AlekseiNikiforovIBM: No, with this change, I can remove commit https://github.com/pytorch/pytorch/pull/120398/commits/2928b2c4d1bfa0ac03881b87cb965432ac381da2

Or I can also drop this change and keep it as just "cpu" for s390x. But for aarch64 "cpu-aarch64" is used, so I did similar "cpu-s390x" for s390x.

.github/workflows/build-manywheel-images.yml
AlekseiNikiforovIBM: No, since I couldn't find where I could download built image for s390x, I just built it manually and backed it into my github actions runner docker image. So it's available locally and isn't needed to be built. I'm also fine with actually building it here and downloading it by actions runner, if it's publicly available somewhere.

huydhn (approved):

Issue #1774, opened 2024-04-05
aarch64 linux: torch.compile performance is 2x slow with nightly torch wheel compared to the wheel built with 'build_aarch64_wheel.py' script


snadampal:
For torchbench benchmarks with dynamo backend, the aarch64 linux nightly wheel performance is 2x slow compared to the wheel I've built using the [pytorch/builder/build_aarch64_wheel.py script for the same pytorch commit.
](https://github.com/pytorch/builder/blob/main/aarch64_linux/build_aarch64_wheel.py)

The difference seems to be coming from 
 the https://github.com/pytorch/builder/blob/main/aarch64_linux/aarch64_ci_build.sh used for nightly builds. I suspect it's with the libomp.

How to reproduce?

```
git clone https://github.com/pytorch/benchmark.git
cd benchmark

# apply this PR: https://github.com/pytorch/benchmark/pull/2187

# setting omp threads =16, because i'm using c7g.4xl instance

OMP_NUM_THREADS=16 python3 run_benchmark.py cpu --model hf_DistilBert --test eval --torchdynamo inductor --freeze_prepack_weights --metrics="latencies,cpu_peak_mem"
```

snadampal:
it's indeed the `libomp.so from conda` in the nightly wheel is causing the issue. I replaced it with `debian libgomp`, the performance is restored, 2x improved.
I'm checking how I can switch to libgomp for the aarch64 nightly and release wheel building.

atalman:
cc @malfet

malfet:
@snadampal so, should we package libomp from debian in our build scripts?

bryantbiggs:
I would love to help get away from these conda packaged deps and instead use something more OS native (i.e. - build in a container using whats provided via apt, yum, dnf, etc.)

snadampal:
The issue is observed for the pytorch 2.3 release candidate wheels as well.
`pip3 install torch==2.3.0  --index-url https://download.pytorch.org/whl/test/cpu`

In the release wheel package, I see we are already packaging both the omp libraries. 
> libgomp-0f9e2209.so.1.0.0 --> from debian
> libomp-b8e5bcfb.so. --> from conda

but looks like all the libraries are linked to `libomp-b8e5bcfb.so`. so, it might be more than packaging, we will have to switch the dependent libraries too.
for experimenting, I renamed ` libgomp-0f9e2209.so.1.0.0 `to `libomp-b8e5bcfb.so` and saw the performance restored.

[...]

PR #1805, opened 2024-04-28
aarch64: cd: switch from libomp to libgomp

snadampal:
In the current version of the CD scripts, torch libraries are linked to llvm openmp because conda openblas-openmp is linked to it. To switch to gnu libgomp, we are building the openblas from sources instead of installing from conda.
Building openBLAS shared library instead of static library to be able to discover LAPACK support in OpenBLAS.

fixes https://github.com/pytorch/builder/issues/1774
cherrypicked https://github.com/pytorch/builder/pull/1803


snadampal:
this is to fix the issue in PyTorch 2.3.x release.
cc: @malfet , @atalman

snadampal:
This PR also fixes: https://github.com/pytorch/pytorch/issues/124922

atalman (approved):

PR #1826, opened 2024-05-21
Add python upgrade guide

atalman:
Create Python upgrade guide

### DIFF
--- a/PYTHON_UPGRADE_GUIDE.MD
+++ b/PYTHON_UPGRADE_GUIDE.MD
@@ -1,17 +1,19 @@
 # Python Upgrade Runbook
 
-So you wanna upgrade PyTorch to support a new Python? Follow these steps in order! They are adapted from previous CUDA upgrade processes.
+So you wanna upgrade PyTorch to support a new Python? Follow these steps in order!
 
 ### 0.Before starting upgrade process. Check the package availability
 
 Package availability to validate before starting upgrade process :
 
-1) Python package is available via python ftp [here](https://www.python.org/ftp/python/) - This is minimal requirement for all wheel builds
+1) Manywheel build: Python version is avalilable in either of these sources:
+- Github: https://github.com/python/cpython
+- Python ftp[here](https://www.python.org/ftp/python/). Please consult the [schedule](https://peps.python.org/pep-0719/)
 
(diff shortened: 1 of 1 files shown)


PYTHON_UPGRADE_GUIDE.MD
```diff
@@ -0,0 +1,36 @@
...
+
+So you wanna upgrade PyTorch to support a new Python? Follow these steps in order! They are adapted from previous CUDA upgrade processes.
+
+### 0.Before starting upgrade process. Check the package availability
+
+Package availability to validate before starting upgrade process :
+
+1) Python package is available via python ftp [here](https://www.python.org/ftp/python/) - This is minimal requirement for all wheel builds
+
+2) Python package is available via anaconda [here](https://anaconda.org/anaconda/python/files) - This is required for conda builds
```
albanD, line 11: We are going to start before this is ready so this is not really a requirement?

albanD (commented):
That sounds good at a high level but I think we should align this with the general process where we expect wheels to be done way ahead. Also build being enabled before tests are enabled.
This doc doesn't have to contain the details of these other components but should most likely contain things like: "wait for Core to fix test" or "wait for core to enable compilation". In between the steps.

PYTHON_UPGRADE_GUIDE.MD
atalman: Agreed. Let me add a the statement about that conda build can be done independently

PYTHON_UPGRADE_GUIDE.MD
```diff
@@ -0,0 +1,38 @@
...
+
+So you wanna upgrade PyTorch to support a new Python? Follow these steps in order!
+
+### 0.Before starting upgrade process. Check the package availability
+
+Package availability to validate before starting upgrade process :
+
+1) Manywheel build: Python version is avalilable in either of these sources:
+- Github: https://github.com/python/cpython
+- Python ftp[here](https://www.python.org/ftp/python/). Please consult the [schedule](https://peps.python.org/pep-0719/)
```
albanD, line 11: nit this pep is the 3.13 schedule only. You can use https://devguide.python.org/versions/ as a long-lived link here.

albanD (approved):
SGTM

PR #1827, opened 2024-05-21
[Manywheel] Add Python-3.13 compilation from branch

atalman:
Same as : https://github.com/pytorch/builder/pull/1427
This however adds an option to build python from latest release branch since we need patch that missed the b1 cut.
cc @albanD @malfet

### DIFF
--- a/common/install_cpython.sh
+++ b/common/install_cpython.sh
@@ -6,7 +6,7 @@ PYTHON_DOWNLOAD_GITHUB_BRANCH=https://github.com/python/cpython/archive/refs/hea
 GET_PIP_URL=https://bootstrap.pypa.io/get-pip.py
 
 # Python versions to be installed in /opt/$VERSION_NO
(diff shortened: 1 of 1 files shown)


common/install_cpython.sh
```diff
@@ -64,7 +65,13 @@ function build_cpython {
     check_var $py_ver
     check_var $PYTHON_DOWNLOAD_URL
     local py_ver_folder=$py_ver
-    wget -q $PYTHON_DOWNLOAD_URL/$py_ver_folder/Python-$py_ver.tgz
+    # Only b2 version of 3.12 is available right now
```
ZainRizvi, line 68: 1. What's a b2 version?
2. Should this be:
```suggestion
    # Only b2 version of 3.13 is available right now
```

ZainRizvi (approved):

common/install_cpython.sh
ZainRizvi: Actually, seems like we're not using the b2 version?

atalman: yes true. The comment is not correct. Let me remove it

malfet (approved):

PR #1861, opened 2024-06-10
[ARM][CUDA][OpenBLAS] Build with NOFORTRAN=1 FLAG

tinglvv:
To fix the below error, need to run `apt install libopenblas-dev`
```
  File "/usr/local/lib/python3.10/dist-packages/torch/__init__.py", line 289, in <module>    from torch._C import *  # noqa: F403
ImportError: libgfortran.so.5: cannot open shared object file: No such file or directory
```

Test build OpenBLAS without gfortran

cc @Aidyn-A @nWEIdia @ptrblck

### DIFF
--- a/common/install_openblas.sh
+++ b/common/install_openblas.sh
@@ -7,7 +7,6 @@ git clone https://github.com/OpenMathLib/OpenBLAS.git -b v0.3.25 --depth 1 --sha
 
 
 OPENBLAS_BUILD_FLAGS="
-NOFORTRAN=1
 NUM_THREADS=128
 USE_OPENMP=1
 NO_SHARED=0
@@ -15,6 +14,11 @@ DYNAMIC_ARCH=1
 TARGET=ARMV8
 CFLAGS=-O3
 "
+if [ "$DESIRED_CUDA" = "cu124" ]; then
+    OPENBLAS_BUILD_FLAGS+="
+NOFORTRAN=1
+"
+fi
 
 OPENBLAS_CHECKOUT_DIR="OpenBLAS"
 
--- a/common/install_openblas.sh
+++ b/common/install_openblas.sh
@@ -14,7 +14,8 @@ DYNAMIC_ARCH=1
 TARGET=ARMV8
 CFLAGS=-O3
 "
-if [ "$DESIRED_CUDA" = "cu124" ]; then
+
+if [ "$GPU_ARCH_TYPE" = "cuda-aarch64" ]; then
     OPENBLAS_BUILD_FLAGS+="
 NOFORTRAN=1
 "
--- a/common/install_openblas.sh
+++ b/common/install_openblas.sh
@@ -7,6 +7,7 @@ git clone https://github.com/OpenMathLib/OpenBLAS.git -b v0.3.25 --depth 1 --sha
 
 
 OPENBLAS_BUILD_FLAGS="
+NOFORTRAN=1
 NUM_THREADS=128
 USE_OPENMP=1
 NO_SHARED=0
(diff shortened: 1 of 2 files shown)


nWEIdia (approved):
LGTM!

common/install_openblas.sh
```diff
@@ -7,6 +7,7 @@ git clone https://github.com/OpenMathLib/OpenBLAS.git -b v0.3.25 --depth 1 --sha
 
 
 OPENBLAS_BUILD_FLAGS="
+NOFORTRAN=1
```
atalman, line 10: This file is called also from [manywheel/Dockerfile_aarch64](https://github.com/pytorch/builder/blob/51d4bfff97cb0089f7c9f5adaab01017681bda14/manywheel/Dockerfile_aarch64#L83) should we instead set this flag in Dockerfile aarch64 cuda ?

nWEIdia: @tinglvv  Please make this script do if else, according to whether Dockerfile called this file with an argument or not. If yes, then it is arm cuda dockerfile calling. Otherwise, it is arm cpu dockerfile calling. Thanks!

common/install_openblas.sh
```diff
@@ -14,6 +14,11 @@ DYNAMIC_ARCH=1
 TARGET=ARMV8
 CFLAGS=-O3
 "
+if [ "$DESIRED_CUDA" = "cu124" ]; then
```
atalman, line 17: I think we need to use BASE_CUDA_VERSION 12.4 here instead 
Looks like DESIRED_CUDA is not set

tinglvv: Thanks for checking, indeed it was not set. Let me use GPU_ARCH_TYPE: cuda-aarch64 instead

manywheel/Dockerfile_cuda_aarch64
```diff
@@ -75,9 +75,10 @@ ADD ./common/install_magma.sh install_magma.sh
 RUN bash ./install_magma.sh ${BASE_CUDA_VERSION} && rm install_magma.sh
 
 FROM base as openblas
+ARG BASE_CUDA_VERSION
 # Install openblas
```
nWEIdia, line 79: Do we even need BASE_CUDA_VERSION? 
We could just 
RUN bash ./install_openblas.sh CUDA_ONLY 

and then in install_openblas.sh, check if $1 is present?

nWEIdia: For ARM CPU build, is BASE_CUDA_VERSION empty or "cpu"?

tinglvv: It may not be worth it to add an extra variable CUDA_ONLY as we could we reuse the BASE_CUDA_VERSION. For CPU build, it will be empty (unset). See https://github.com/pytorch/builder/blob/main/manywheel/build_docker.sh#L36.

nWEIdia: It is not a variable, just an argument, string. 

The purpose of BASE_CUDA_VERSION is to just supply a string, right? 

Either way it works.

tinglvv: BASE_CUDA_VERSION will be supplied only when CUDA arm build is running, so I think it might be reasonable to use.

atalman:
hi @tinglvv yes this looks good. I do see it here on gpu aarch64 builds:
https://github.com/pytorch/builder/actions/runs/9519461690/job/26242701970?pr=1861#step:5:1221

And not here:
https://github.com/pytorch/builder/actions/runs/9519461690/job/26242706088?pr=1861#step:5:686

atalman (approved):

PR #1879, opened 2024-06-21
Add xpu manylinux 2.28 wheel build

chuanqi129:


manywheel/build_xpu.sh
```diff
@@ -0,0 +1,61 @@
...
+DEPS_LIST=(
+    "$LIBGOMP_PATH"
+)
+
+DEPS_SONAME=(
+    "libgomp.so.1"
+)
+
+rm -rf /usr/local/cuda*
+source /opt/intel/oneapi/pytorch-gpu-dev-0.5/oneapi-vars.sh
```
atalman, line 53: This file looks almost same as ``build_cpu.sh`` except fot this line and ``wheelhousexpu``. Perhaps it possible to reuse ``build_cpu.sh`` in this case rather then repeating all the code ?

chuanqi129: Thanks @atalman, yes, good suggestion. But I am worried that in the future these two will be separated, and it will be clearer if they are separated. How do you think? I'm also ok to reuse `build_cpu.sh` if we prefer this method in current stage.

atalman (approved):
lgtm

PR #1896, opened 2024-06-28
Switch mkl-static and mkl-include install to pip

atalman:
Conda packages are missing from ``anaconda.org/intel`` hence loading them from pypi


huydhn (approved):

atalman:
will reland here: https://github.com/pytorch/builder/pull/1899

PR #1900, opened 2024-07-01
S390x update

AlekseiNikiforovIBM:
Current state of s390x builder image.

Switch to manylinux instead of ubuntu.

When no python version is specified, python 3.12 is used.


AlekseiNikiforovIBM:
Abandoning this change since updated files moved to pytorch repository.

PR #1904, opened 2024-07-02
Patch mirrorlist.centos.org no longer online

atalman:
See following: https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve

Here is the failure:
https://github.com/pytorch/builder/actions/runs/9716034788/job/26898991378#step:4:80


ajindal1:
@atalman I believe we need to add this to [release/2.4](https://github.com/pytorch/builder/tree/release/2.4) branch.

atalman:
HI @ajindal1 we pin docker images for the release here: https://github.com/pytorch/pytorch/pull/128347/files so this issue have not manifested itself on the release.

PR #1906, opened 2024-07-02
Add smoke test for CUDA_VISIBLE_DEVICES

atalman:
Add smoke test for : https://github.com/pytorch/pytorch/issues/128819

### DIFF
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -190,8 +190,7 @@ def smoke_test_cuda(package: str, runtime_error_check: str, torch_compile_check:
 
         if runtime_error_check == "enabled":
             test_cuda_runtime_errors_captured()
-
-       
+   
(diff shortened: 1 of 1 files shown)


jeanschmidt (approved):

test/smoke_test/smoke_test.py
```diff
@@ -182,6 +182,12 @@ def smoke_test_cuda(package: str, runtime_error_check: str, torch_compile_check:
             print(f"torch nccl version: {torch.cuda.nccl.version()}")
 
 
+         # test settind CUDA_VISIBLE_DEVICES
```
kit1980, line 185: typo "settind"

malfet:
Calling `CUDA_VISIBLE_DEVICES` after first allocation is done would not have any effect on the system, so it does not test for the regression, is it?

PR #1925, opened 2024-07-18
Pytorch-conda 11.8 build include cuda-nvcc

atalman:
cuda-nvcc includes ptxas this is required to make torch.compile to work with older CUDA driver.
Remove older 11.7 package build


malfet (approved):
Sure, though I would appreciate a test plan of sorts. `cuda-nvcc` package is huge and not needed by majority of the devs. If you just want PTXAS, then we should bundle it during the build, rather than doing something like that

PR #1930, opened 2024-07-23
build aarch64 linux cpu wheel with the linker script optimization

snadampal:
the prioritized symbols linking is giving more than 10% performance uplift for inference across several models.


nWEIdia (approved):
Nice!

atalman (approved):

PR #1932, opened 2024-07-24
add +PTX to CUDA 12.4 nightly binaries

ptrblck:
This PR adds `+PTX` back to our x86 CUDA 12.4 binaries, which was previously dropped.
We should consider adding this change only to nightly binaries and keep versioned releases as they are to keep the binary size lower.
However, I'm unsure how and where the nightly builds are calling this script.

CC @malfet

### DIFF
--- a/manywheel/build_cuda.sh
+++ b/manywheel/build_cuda.sh
@@ -63,7 +63,7 @@ case ${CUDA_VERSION} in
         if [[ "$GPU_ARCH_TYPE" = "cuda-aarch64" ]]; then
             TORCH_CUDA_ARCH_LIST="9.0"
         else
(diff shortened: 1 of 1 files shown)


atalman:
@ptrblck what is the size increase for these binaries ?

ptrblck:
CC @nWEIdia do you know what the approx. size increase for `+PTX` would be? I think you just recently measured it.

nWEIdia:
> CC @nWEIdia do you know what the approx. size increase for `+PTX` would be? I think you just recently measured it.

The approximate size increase for `+PTX` would be 200MB.

nWEIdia (approved):
LGTM!

malfet (approved):

malfet:
Let's give it a try, we can always rollback later

PR #1952, opened 2024-08-08
Check that CPUInductor selects valid SIMD ISA

malfet:
I.e. for every platform we support, it should be something but invalid AVX512 on Intel Linux, and NEON on Apple silicon and Linux aarch64

2.4 aarch64 failures are expected, as https://github.com/pytorch/pytorch/pull/129075 has not been picked for the release yet


xuhancn:
Hi @malfet 

Current CPU isa code was moved to https://github.com/pytorch/pytorch/blob/main/torch/_inductor/cpu_vec_isa.py, if we need to update code for this change?

PR #1987, opened 2024-09-12
Fix xpu env setup for binaries validate script

chuanqi129:
Inherit XPU env from `check_binary.sh` to `.github/scripts/validate_binaries.sh`


atalman (approved):
lgtm

PR #1999, opened 2024-09-27
[Validations] Add option to use meta cdn for pypi index

atalman:


tinsingcheong (approved):
Stamp to unblock, just a quick question: how do we know this is only for nightly build? I didn't see specific condition to enable use_meta_cdn or maybe that's in another diff?

PR #2000, opened 2024-10-01
Use python 3.9 for libtorch builds by default

atalman:
Part of deprecation of Python 3.8


kit1980 (approved):

albanD (approved):
Aren't libtorch wheels supposed to not have python?
Or maybe I'm confused what libtorch wheels contain? :p

PR #2002, opened 2024-10-01
Enable xpu windows wheel smoke test

chuanqi129:
Works for https://github.com/pytorch/pytorch/issues/114850

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -6,6 +6,10 @@ else
     export PYTHON_RUN="python3"
     if [[ ${TARGET_OS} == 'windows' ]]; then
         export PYTHON_RUN="python"
+        # Currently xpu env need a helper script to activate
+        if [[ ${DESIRED_CUDA} == "xpu" ]]; then
+            export PYTHON_RUN="${PWD}/.github/scripts/xpu_env_helper.bat python"
+        fi
     fi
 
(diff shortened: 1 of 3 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -6,6 +6,10 @@ else
     export PYTHON_RUN="python3"
     if [[ ${TARGET_OS} == 'windows' ]]; then
         export PYTHON_RUN="python"
+        # Currently xpu env need a helper script to activate
+        if [[ ${DESIRED_CUDA} == "xpu" ]]; then
```
atalman, line 10:
```suggestion
        if [[ ${MATRIX_GPU_ARCH_TYPE} == "xpu" ]]; then
```

atalman (approved):
lgtm

.github/scripts/xpu_env_helper.bat
```diff
@@ -0,0 +1,17 @@
...
+shift
+goto start
+
+:done
+if "%args%" == "" (
+    echo Usage: xpu_env_helper.bat [command] [args]
+    echo e.g. xpu_env_helper.bat icpx --version
+)
+
+%args% || exit /b 1
```
malfet, line 17: What were the reason for using Batch script? Why can't it be a power shell one?

atalman: cc @chuanqi129

PR #2009, opened 2024-10-08
[Release 2.5] Align xpu windows wheel build env

chuanqi129:
Cherry-pick from https://github.com/pytorch/builder/pull/1988


malfet:
Not a regression, therefore not suitable for 2.5.1. If next patch release will be planned, we should reconsider

PR #2026, opened 2024-10-31
Build magma without conda and push tarball

afrittoli:
This is the first step towards removing `conda` from Magma builds. 

This still relies on the python conda image to build, as that image already contains the required dependencies and various versions of cuda.
Use a build script to build magma, instead of conda-build, to produce the binary tarball without the rest of the conda package. An alternative approach could be to produce a no-python wheel, which we could consider if we needed to maintain multi-arch versions of this.

Upload the package to downloads.pytorch.org instead of the anaconda pytorch channel (TBD).

This will require updating the "install_conda.sh" script so that magma can be installed directly from that tarball.


afrittoli:
superseded by https://github.com/pytorch/pytorch/pull/139888

PR #2033, opened 2024-11-07
Update magma artifact upload

atalman:
Currently only 1 artifact is getting uploaded. I see this in the log: 
```
Run actions/upload-artifact@v4
  with:
    path: magma_*_cuda*_*.7z
    overwrite: true
    name: artifact
    if-no-files-found: warn
    compression-level: [6](https://github.com/pytorch/builder/actions/runs/11713319173/job/32625864948#step:6:6)
    include-hidden-files: false
```

This should enable all artifacts to be uploaded

### DIFF
--- a/.github/workflows/build-magma-windows.yml
+++ b/.github/workflows/build-magma-windows.yml
@@ -36,6 +36,7 @@ jobs:
         with:
           path: magma_*_cuda*_*.7z
           overwrite: true
+          name: artifact_${{ matrix.cuda_version }}_${{ matrix.config }}
   push-windows-magma:
     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     environment: magma


afrittoli (approved):

.github/workflows/build-magma-windows.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         with:
           path: magma_*_cuda*_*.7z
           overwrite: true
+          name: artifact_${{ matrix.cuda_version }}_${{ matrix.config }}
```
afrittoli, line 39: With this, the artifact is not uploaded to a zip with a different name:

```
2024-11-07T15:36:52.7793012Z Artifact artifact_126_Release.zip successfully finalized. Artifact ID 2158495441
2024-11-07T15:36:52.7794688Z Artifact artifact_126_Release has been successfully uploaded! Final size is 62089698 bytes. Artifact ID is 2158495441
2024-11-07T15:36:52.7804031Z Artifact download URL: https://github.com/pytorch/builder/actions/runs/11724637191/artifacts/2158495441
```

That's fine, but how does that help?
The push job downloads the artifact and expects something called `magma_*_cuda*.7z` - would that still work?

atalman: Yes this is upload to GHA. So that artifact can be downloaded from the workflow page

